PR #19, opened 2020-07-14
resolved the ambiguity between types and primitives.hpp

### LINKED ISSUE #17: primitives.hpp and types.hpp ambiguity
Unless there is some good reason to keep them seperate I advise on joining these two

For which one please discuss in the comments or via discord

Algo-ryth-mix:
fixes #17 
- rename *sfinae.hpp to *meta.hpp
- moved contents from types.hpp to primitives.hpp
- made types.hpp the single include for primitives and meta


Algo-ryth-mix:
- pr from fork breakes pipeline, as GITHUB_TOKEN is wrong, this needs to be investigated, for now the fix is to push the change to the base repo directly

PR #25, opened 2020-07-15
cppcheck

Algo-ryth-mix:
integrate cppcheck into the workflow


Algo-ryth-mix:
this is sadly unsuitable as it can only run on pull and not on pr

PR #35, opened 2020-07-17
add sparse_map

GlynLeine:
[Add] sparse_map.hpp

### DIFF
--- a/applications/test_game/source.cpp
+++ b/applications/test_game/source.cpp
@@ -1,11 +1,85 @@
 #include <iostream>
+
 #define ARGS_ENTRY
 #include <core/core.hpp>
 
 #include "module/testModule.hpp"
 
+struct sah
+{
+	std::string value;
+};
+
 void ARGS_CCONV reportModules(args::core::Engine* engine)
 {
 	std::cout << "Hello Args!" << std::endl;
 	engine->reportModule<TestModule>();
+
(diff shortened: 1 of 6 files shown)


applications/test_game/source.cpp
```diff
@@ -1,11 +1,85 @@
...
+
 void ARGS_CCONV reportModules(args::core::Engine* engine)
 {
 	std::cout << "Hello Args!" << std::endl;
 	engine->reportModule<TestModule>();
+
+	std::cout << "_____________________________________________________" << std::endl;
+
+
+	args::core::sparse_map<int, sah> testMap;
```
Algo-ryth-mix, line 21: shouldn't a sparse set only hold uint ? 🤔

Algo-ryth-mix (approved):

applications/test_game/source.cpp
GlynLeine: normally yes, but by using a hash map as the sparse container and storing a pair with the key and the value in the dense container we can map any type value to any other type value and using the concept of a sparse set inbetween

GlynLeine: the traditional sparse_set is still planned and will come either tomorrow or monday prolly

Algo-ryth-mix: fine by me. there is an implementation of sparse_set in lpe if you want a pretty much ready made one

PR #36, opened 2020-07-20
sparse set implementation

GlynLeine:


args/core/containers/sparse_set.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+	private:
+		dense_container m_dense;
+		sparse_container m_sparse;
+
+		size_type m_size = 0;
+		size_type m_capacity = 0;
+
+	public:
+		A_NODISCARD inline iterator begin() { return m_dense.begin(); }
```
Algo-ryth-mix: the `inline`'s here are redundant and should be removed, the specifier is only required when actually implementing the function

args/core/containers/sparse_map.hpp
```diff
@@ -53,37 +53,37 @@ namespace args::core
 		size_type m_capacity = 0;
 
 	public:
-		A_NODISCARD iterator begin() { return m_dense.begin(); }
-		A_NODISCARD const_iterator begin() const { return m_dense.cbegin(); }
+		A_NODISCARD inline iterator begin() { return m_dense.begin(); }
```
Algo-ryth-mix: see the review on `sparse_set.hpp`

Algo-ryth-mix (changes requested):

args/core/containers/sparse_set.hpp
Algo-ryth-mix: and or when not in a class definition

Algo-ryth-mix (approved):

PR #40, opened 2020-07-26
Feature/component storage

GlynLeine:


applications/test_game/source.cpp
```diff
@@ -9,156 +9,96 @@ using namespace args::core;
 
 struct sah
```
Algo-ryth-mix, line 10: what is `sah anyways?`

Algo-ryth-mix (approved):

Algo-ryth-mix:
I will ignore this glitch in the matrix for now and fix the build-action tomorrow

applications/test_game/source.cpp
GlynLeine: sah is just a temporary name i use for stuff, it's short for whatssup

PR #48, opened 2020-07-30
Feature/entity querying

GlynLeine:
entity querying plus tbh loads of other minor things...


GlynLeine:
we have static thread_local storage, and i don't rly know how to define it without linker errors...

GlynLeine (commented):
[commit #88f24a2](https://github.com/Args-Engine/Args-Engine/pull/48/commits/88f24a2fec32146d8ff1150ff4efd2aed2f1be86)
@Algo-ryth-mix can you have a look at these changes?

args/core/async/readonly_rw_spinlock.cpp
```diff
@@ -0,0 +1,20 @@
+#include <core/async/readonly_rw_spinlock.hpp>
+
+namespace args::core::async
+{
+	std::atomic_uint readonly_rw_spinlock::lastId;
+
+	std::unordered_map<uint, int>& readonly_rw_spinlock::localStates()
+	{
+		static thread_local std::unordered_map<uint, int> states;
```
Algo-ryth-mix, line 9: considering that states is one of `idle` ,`read` or `write` as far as I understand it, an `std::unordered_map<>` appears overkill for me, consider replacing it with an `std::array<>` maybe ?

args/core/async/transferable_atomic.hpp
```diff
@@ -1,21 +1,34 @@
...
 	public:
 		transferable_atomic() noexcept = default;
 
-		constexpr transferable_atomic(T val) noexcept : m_atomic(val) {}
+		constexpr transferable_atomic(T val) noexcept : m_atomic(val), m_lock() {}
+
+		transferable_atomic(const std::atomic<T>& other) : m_atomic(other.load(std::memory_order_acquire)), m_lock() {}
 
-		transferable_atomic(const std::atomic<T>& other) : m_atomic(other.load(std::memory_order_acquire)) {}
+		transferable_atomic(const transferable_atomic<T>& other) : m_atomic(other->load(std::memory_order_acquire)), m_lock() {}
```
Algo-ryth-mix, line 26: considering that it implements a copy-ctor, and the type is quite literally called `transferable_atomic` you might want to also implement a move-constructor, a copy assignment operator, a move assignment operator and a destructor.
(see cppcoreguidelines-special-member-functions)
when these do not apply, explicitly mark them as `= delete;`

args/core/containers/hashed_sparse_set.hpp
```diff
@@ -0,0 +1,345 @@
...
+		A_NODISCARD iterator begin() { return m_dense.begin(); }
+		A_NODISCARD const_iterator begin() const { return m_dense.cbegin(); }
+
+		A_NODISCARD iterator end() { return m_dense.begin() + m_size; }
+		A_NODISCARD const_iterator end() const { return m_dense.cbegin() + m_size; }
+
+		/**@brief Returns the amount of items in the sparse_map.
+		 * @returns size_type Current amount of items contained in sparse_map.
+		 */
+		A_NODISCARD size_type size() const noexcept { return m_size; }
```
Algo-ryth-mix, line 64: considering you implement `size()` and `capacity()` you should also think about `max_size()`

args/core/containers/hashed_sparse_set.hpp
```diff
@@ -0,0 +1,345 @@
...
+		sparse_container m_sparse;
+
+		size_type m_size = 0;
+		size_type m_capacity = 0;
+
+	public:
+		A_NODISCARD dense_container& dense() { return m_dense; }
+		A_NODISCARD const dense_container& dense() const { return m_dense; }
+
+		A_NODISCARD iterator begin() { return m_dense.begin(); }
```
Algo-ryth-mix, line 53: to adhere to the standard you should think about also implementing `cbegin()` and `cend()`next to `begin() const` and `end() const`

args/core/containers/hashed_sparse_set.hpp
```diff
@@ -0,0 +1,345 @@
...
+#pragma endregion
+
+#pragma region operator[]
+		/**@brief Returns item from dense container.
+		 * @param index Index of item in dense container.
+		 */
+		A_NODISCARD value_reference operator[](size_type&& index)
+		{
+			if (index < 0 || index > m_size)
+				throw std::out_of_range("Index out of range.");
```
Algo-ryth-mix: "hashed_sparse_set subscript out of range", would imo be a better exception

args/core/types/type_util.hpp
```diff
@@ -1,13 +1,25 @@
...
  */
 
 namespace args::core
 {
+	/**@brief Forcefully casts non pointer/reference type value from one to another.
+	*/
+	template<typename T, typename U>
+	T force_value_cast(U value)
+	{
+		static_assert(alignof(T) == alignof(U), "Illegal cast of unalligned types. Approaching undefined behaviour.");
```
Algo-ryth-mix: Approaching ?

args/core/types/type_util.hpp
```diff
@@ -106,11 +86,11 @@ namespace args::core
 		hash = 0x811c9dc5;
 		uint32 prime = 0x1000193;
 		cstring name = typeName<T>();
-		for (int i = 0; i < std::strlen(name); i++)
-		{
-			byte_t value = name[i];
-			hash = hash ^ value;
+		while(*name != '\0')
```
Algo-ryth-mix: note that this is potentially a very dangerous approach, consider only allowing string literal hashing via `auto operator"_hash(const char *, std::size_t)` to avoid potential arbitray memory lookup. (consider what happens when someone malicious injects a non null terminated string!)
use std::string::size() on normal strings or replace it with std::string_view::size() to cover both const char* and std::string simultaneously. 
```suggestion
		for(std::string::size_type i = 0; i < name.size(); ++i) 
		//or
		for(const char n : name)
```

args/core/types/type_util.cpp
```diff
@@ -0,0 +1,36 @@
+#include <core/types/type_util.hpp>
+
+namespace args::core
+{
+	id_type ARGS_FUNC nameHash(cstring name)
```
Algo-ryth-mix, line 5: consider creating a constexpr version with a crc32 lut
(minor suggestion, might improve speeds a bit :) )

args/core/platform/platform.hpp
```diff
@@ -9,8 +9,45 @@
...
+	 */
+	#define ARGS_DEBUG
+#else
+	/**@def ARGS_RELEASE
+	 * @brief Defined in release mode.
+	 */
+	#define ARGS_RELEASE
+#endif
+
+#if (!defined(ARGS_LOW_POWER) && !defined(ARGS_HIGH_PERFORMANCE))
```
Algo-ryth-mix, line 24: what is the usecase for ARGS_LOW_POWER ?

args/core/platform/platform.hpp
```diff
@@ -9,8 +9,45 @@
  */
 #define ARGS_CPP17V 201703L
 
+#if defined(_DEBUG)
```
Algo-ryth-mix: note that on linux compilers it is common to define DEBUG and NDEBUG instead of _DEBUG. (NDEBUG=Release).
```suggestion
#if defined(_DEBUG) || defined(DEBUG)
```

args/core/entry/entry_point.hpp
```diff
@@ -1,34 +1,91 @@
...
- * @note When not using ARGS_ENTRY you must call creation and initialisation of the engine manually.
+ * @note When not using ARGS_ENTRY you must call creation and initialization of the engine manually.
  */
 
 
-/**@brief Reports engine modules to the engine, must be implemented by you.
- * @param [in] engine The engine object as ptr *
- * @ref args::core::Engine::reportModule<T,...>()
- */
+ /**@brief Reports engine modules to the engine, must be implemented by you.
```
Algo-ryth-mix, line 16: some indentation errors ?

args/core/entry/entry_point.hpp
```diff
@@ -1,34 +1,91 @@
...
+
+		try
+		{
+			args::core::Engine engine;
+			reportModules(&engine);
+
+			engine.init();
+
+	#if defined(ARGS_DEBUG)
+			std::cin.ignore().get();
```
Algo-ryth-mix, line 52: consider replacing this with
```suggestion
#if defined(ARGS_WINDOWS)
	_getch();
#else
   	char c = 0;
    system("/bin/stty raw");
    while (c == 0) {
        c = getchar();
    }
    system("/bin/stty cooked");
    return c;
#endif	
```
for a better experience ("press any key" instead of "enter a single char + enter")

requires \<cstdlib\> and
on windows:
	\<conio.h\>
on everything else
	\<unistd.h\>
	\<cstdio\>

args/core/ecs/queryregistry.cpp
```diff
@@ -0,0 +1,201 @@
...
+		async::readonly_guard refguard(m_referenceLock);
+
+		if (queryId == invalid_id || !m_references.contains(queryId))
+			return;
+
+		size_type& referenceCount = m_references.get(queryId);
+		referenceCount--;
+		if (referenceCount == 0)
+		{
+			async::readwrite_guard refWguard(m_referenceLock); // permitted to double lock, elevates permission to write.
```
Algo-ryth-mix, line 184: this seems like a potential deadlock, related #49 ?

args/core/ecs/queryregistry.cpp
```diff
@@ -0,0 +1,201 @@
+#include <core/ecs/queryregistry.hpp>
+#include <core/ecs/ecsregistry.hpp>
+#include <core/ecs/entity_handle.hpp>
+#include <core/ecs/entityquery.hpp>
+#include <algorithm>
+
```
Algo-ryth-mix, line 6: lots of comment-worthy functions

args/core/ecs/entityquery.cpp
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	inline void EntityQuery::addComponentType(id_type componentTypeId)
+	{
+		hashed_sparse_set<id_type> componentTypes;
+		if (m_id)
+			componentTypes = m_registry.getComponentTypes(m_id);
+
+		componentTypes.insert(componentTypeId);
+
```
Algo-ryth-mix, line 31: not super sure what is going on here

args/core/ecs/ecsregistry.cpp
```diff
@@ -1,71 +1,206 @@
...
+
+		entity_data* data = nullptr;
+
+		{
+			async::readonly_guard guard(m_entityDataLock);
+			data = &m_entityData[entityId];
+		}
+
+		{
+			async::readonly_guard guard(m_familyLock); // Possibly deadlocks? Will also lock atomic_sparse_map::m_container_lock for the family.
```
Algo-ryth-mix, line 145: if it is a potential deadlock , please mark with TODO(\<assignee\>): to figure out

args/core/containers/sparse_set.hpp
```diff
@@ -222,6 +261,8 @@ namespace args::core
 		 */
 		A_NODISCARD reference operator[](size_type&& index)
 		{
+			if (index < 0 || index > m_size)
+				throw std::out_of_range("Index out of range.");
```
Algo-ryth-mix: "sparse_set subscript out of range" ?

args/core/containers/sparse_map.hpp
```diff
@@ -154,8 +160,64 @@ namespace args::core
...
+		 */
+		A_NODISCARD bool equals(self_const_reference other) const
+		{
+			if (m_size == other.m_size)
+			{
+				bool equal = true;
+				for (int i = 0; i < m_size; i++)
+					equal = equal && other.contains(m_dense_key[i]) && get(m_dense_key[i]) == other.get(m_dense_key[i]);
+
+				return equal;
```
Algo-ryth-mix: there appears to be an early out option:
```suggestion
				for (int i = 0; i < m_size; ++i)
					if(!(other.contains(m_dense_key[i]) &&
					     get(m_dense_key[i]) == other.get(m_dense_key[i])))
						return false;
						
				return true;
```

args/core/containers/sparse_map.hpp
```diff
@@ -154,8 +160,64 @@ namespace args::core
...
+				return true;
+
+			if (m_size == 0 || m_size < other.m_size)
+				return false;
+
+			bool overlap = true;
+			for (int i = 0; i < other.m_size; i++)
+				overlap = overlap && contains(other.m_dense_key[i]);
+
+			return overlap;
```
Algo-ryth-mix: see below

Algo-ryth-mix (changes requested):

args/core/async/readonly_rw_spinlock.cpp
Algo-ryth-mix: nvm I see how this works. depending on the number of threads it still might be better to have a different data structure.

Algo-ryth-mix (approved):

Algo-ryth-mix (changes requested):
spurious review

Algo-ryth-mix (commented):
new additions look good

args/core/ecs/queryregistry.cpp
Algo-ryth-mix: related to #50, resolve via extra pr

Algo-ryth-mix (approved):

Algo-ryth-mix:
this is unblocked now, but still consider implementing the features/bugfixes mentioned above

PR #63, opened 2020-08-03
Bugfix/move and reference use

GlynLeine:


GlynLeine:
multiguard PR should probably be merged first!

args/core/containers/sparse_map.hpp
```diff
@@ -420,14 +421,15 @@ namespace args::core
...
+				k = *itr_key;
+				m_sparse[k] = m_size;
 				++m_size;
 			}
+			else
+				k = key;
 
-
-			return m_dense_value[m_sparse[key]];
+			return m_dense_value[m_sparse[k]];
```
Algo-ryth-mix: mind explaining a bit what is going on here ?

args/core/async/readonly_rw_spinlock.hpp
```diff
@@ -161,26 +395,137 @@ namespace args::core::async
...
-				lock.readState.store(readonly_rw_spinlock::read, std::memory_order_release);
+				locked = true;
+				for (uint i : unlockedLocks)
+				{
+					if (m_locks[i]->try_lock(write))
+						unlockedLocks.erase(i);
+					else
+						locked = false;
+				}
+			} while (!locked);
```
Algo-ryth-mix: how does this actually prevent deadlocks ? shouldn't all locks be unlocked on an failed attempt in order to not spin endlessly ?

Algo-ryth-mix (changes requested):

args/core/containers/sparse_map.hpp
Algo-ryth-mix: (in the form of comments preferably)

GlynLeine: uhm sure, but it's quite standard. using a move operation might invalidate the previous rvalue or lvalue ref you had. thus you cannot use that item anymore.

args/core/async/readonly_rw_spinlock.hpp
GlynLeine: yes you're right...

args/core/containers/sparse_map.hpp
Algo-ryth-mix: yes, but being quite standard does not mean it is particularly readable as a whole
```cpp
		value_reference operator[](key_type&& key)
		{
			key_type k;
			if (!contains(key))
			{
				if (m_size >= m_capacity)
					reserve(m_size + 1);

				auto itr_value = m_dense_value.begin() + m_size;
				*itr_value = std::forward<value_type>(value_type());

				auto itr_key = m_dense_key.begin() + m_size;
				*itr_key = std::move(key);
				k = *itr_key;
				m_sparse[k] = m_size;
				++m_size;
			}
			else
				k = key;

			return m_dense_value[m_sparse[k]];
		}
```
here is the piece of code in question without github diff stuff around it. Not super readbale imo.
It does not help that we have itr_key, key, and k here which are seamingly the same things (and all quite terrible names for variables)

args/core/async/readonly_rw_spinlock.hpp
```diff
@@ -324,35 +324,51 @@ namespace args::core::async
...
 	public:
 		/**@brief Creates readonly multi-guard and locks for Read-only.
 		 */
 		template<typename lock_type1 = readonly_rw_spinlock, typename lock_type2 = readonly_rw_spinlock, typename... lock_typesN>
 		readonly_multiguard(lock_type1& lock1, lock_type2& lock2, lock_typesN&... locks)
 		{
-			m_locks.push_back(&lock1);
-			m_locks.push_back(&lock2);
-			(m_locks.push_back(&locks), ...);
+			fill<sizeof...(locks) + 2>(lock1, lock2, locks...);
```
Algo-ryth-mix: why even use fill here ?
```suggestion
			m_locks = {{ &lock1, &lock2, &locks... }};
```
longer example with better optimization as well https://godbolt.org/z/rWx7vx

args/core/containers/sparse_map.hpp
GlynLeine: Untitled

is this okay?

args/core/ecs/entityquery.cpp
```diff
@@ -25,7 +25,7 @@ namespace args::core::ecs
 	{
 		hashed_sparse_set<id_type> componentTypes;
 		if (m_id)
-			componentTypes = m_registry.getComponentTypes(m_id);
+			componentTypes = m_registry.getComponentTypes(m_id); /// comment this glyn.
```
Algo-ryth-mix, line 28: I heard that talking to yourself is not healthy, I am unsure about writing yourself.

Algo-ryth-mix (approved):

args/core/containers/sparse_map.hpp
Algo-ryth-mix: yes

args/core/async/readonly_rw_spinlock.hpp
GlynLeine: got it

PR #64, opened 2020-08-04
Bugfix/comments

GlynLeine:


args/core/engine/module.hpp
```diff
@@ -27,6 +27,8 @@ namespace args::core
 		 * @ref ARGS_PURE
 		 */
 		virtual void init() ARGS_PURE;
+
+		virtual ~Module() ARGS_IMPURE;
```
Algo-ryth-mix, line 31: this should really be marked = default;

Algo-ryth-mix (changes requested):

args/core/engine/module.hpp
GlynLeine: resolution can be found in commit #59333b4

PR #65, opened 2020-08-04
Feature/clock and time

GlynLeine:


args/core/time/time_span.hpp
```diff
@@ -0,0 +1,30 @@
...
+{
+	template<typename time_type>
+	class time_span
+	{
+	private:
+		std::chrono::duration<time_type> m_duration;
+
+	public:
+		time_span(std::chrono::duration<time_type> duration) : m_duration(duration) {}
+
```
Algo-ryth-mix, line 15: as per cppreference.org I require 
`microfortnights`(std::ratio<12096,10000>) and
`nanocenturies` (std::ratio<3155,1000>)

Algo-ryth-mix (approved):

PR #69, opened 2020-08-06
Feature/scheduling

GlynLeine:


args/core/async/ring_sync_lock.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+		uint waiterCount()
+		{
+			readonly_guard guard(m_waitersLock);
+			return m_waiters[m_rank.load(std::memory_order_acquire)]->load(std::memory_order_acquire);
+		}
+
+		std::thread::id ownerThread() { return owningThread; }
+	};
+}
```
Algo-ryth-mix: what does any of this do ?

args/core/engine/system.hpp
```diff
@@ -0,0 +1,64 @@
...
+			process->setOperation(operation);
+			m_processes.insert(id, std::move(process));
+
+			m_scheduler->hookProcess(chainName, m_processes[id].get());
+		}
+
+	public:
+		System() : SystemBase(typeHash<SelfType>(), undecoratedTypeName<SelfType>()) {}
+		virtual ~System() = default;
+	};
```
Algo-ryth-mix, line 63: no comments ?

args/core/scheduling/scheduler.hpp
```diff
@@ -0,0 +1,402 @@
...
+						m_processChains.erase(id);
+				}
+
+				if (m_localChain.id())
+					m_localChain.runInCurrentThread();
+
+				if (syncRequested())
+					waitForProcessSync();
+			}
+		}
```
Algo-ryth-mix: header only is neat and all, but I would move this to a .cpp for readability

Algo-ryth-mix (changes requested):

args/core/scheduling/scheduler.hpp
GlynLeine: well you asked before on a different class if ARGS_API was rly necessary, and since it's not rly necessary here i thought i'd avoid using ARGS_API

args/core/engine/system.hpp
GlynLeine: I didn't comment anything yet because i only have a couple of hours left to implement an event system as well.
tomorrow i'm leaving to france and i'll be without internet for about a week.

Algo-ryth-mix (approved):

PR #73, opened 2020-08-13
Feature/math

Algo-ryth-mix:
initial integration of some maths. Most of it is just glm, there is the custom additions of 
math/constants.hpp and math/trigonometry.hpp glm is wrapped in args::core::math


GlynLeine:
all checks failed?

Algo-ryth-mix:
anaylze-action is still broken from event

GlynLeine (approved):

PR #74, opened 2020-08-13
added tools/cppcheck2github.py

Algo-ryth-mix:
added a custom tool to reformat cppcheck errors as github messages:
https://github.com/Algo-ryth-mix/cppcheck-github
Formats errors like this
grafik
instead of this 
grafik

Note that this also changes the return type to always be 0, if wanted we can change this back in the script

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -33,4 +33,4 @@ jobs:
        /home/runner/.local/bin/analyze-build --status-bugs
 
     - name: cppcheck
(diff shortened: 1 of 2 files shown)


GlynLeine:
what's the cppcheck syntax error?

Algo-ryth-mix:
that is the template thing we discovered a few days ago.
I've raised the issue here https://sourceforge.net/p/cppcheck/discussion/general/thread/91801de693/ but I am yet to receive an answer
grafik

GlynLeine (approved):

PR #81, opened 2020-08-25
window/application context

GlynLeine:
still needs comments and docs


GlynLeine:
idk how to get rid of those warnings from glm with cppcheck... @Algo-ryth-mix

Algo-ryth-mix (approved):

PR #106, opened 2020-09-08
Feature/physics naming convention fix

Developer-The-Great:


Algo-ryth-mix (changes requested):
code-conventions adaptions look passable, needs more comments tho

GlynLeine (changes requested):
Some files still use upper case letters and are missing _ underscores

GlynLeine (approved):

Algo-ryth-mix (approved):

PR #107, opened 2020-09-08
added spdlog logging framework, please use instead of cout

Algo-ryth-mix:


args/core/engine/engine.hpp
```diff
@@ -12,6 +12,8 @@
 #include <vector>
 #include <memory>
 
+#include "core/logging/logging.hpp"
```
GlynLeine: use <>

args/core/logging/logging.cpp
```diff
@@ -0,0 +1,9 @@
+#include "logging.hpp"
```
GlynLeine: use<>

args/core/logging/logging.hpp
```diff
@@ -0,0 +1,153 @@
+#pragma once
+
+#define SPDLOG_HEADER_ONLY
+#include "spdlog/spdlog.h"
```
GlynLeine: use <>

GlynLeine (changes requested):

GlynLeine (approved):

PR #119, opened 2020-09-15
Materials, jobs, thread commands, fix spinlock, fix projection

GlynLeine:


applications/test_game/systems/testsystem.hpp
```diff
@@ -52,23 +54,36 @@ class TestSystem final : public System<TestSystem>
...
+        rendering::material_handle vertexH;
+
+        {
+            async::readwrite_guard guard(*window.lock);
+            app::ContextHelper::makeContextCurrent(window);
 
-        bindToEvent<events::component_creation<app::window>, &TestSystem::disableCursor>();
+            modelH = rendering::ModelCache::create_model("test", "basic://models/Cube.obj"_view);
+            wireframeH = rendering::MaterialCache::create_material("wireframe", "basic:/shaders/wireframe.glsl"_view);
+            vertexH = rendering::MaterialCache::create_material("vertex", "basic:/shaders/position.glsl"_view);
```
Algo-ryth-mix: `basic:/`vs `basic://` ?

args/core/scheduling/scheduler.cpp
```diff
@@ -6,13 +6,74 @@ namespace args::core::scheduling
...
+    std::queue<Scheduler::runnable> Scheduler::m_jobs;
+    sparse_map<std::thread::id, async::readonly_rw_spinlock> Scheduler::m_commandLocks;
+    sparse_map<std::thread::id, std::queue<Scheduler::runnable>> Scheduler::m_commands;
+
+    void Scheduler::threadMain(bool* exit, bool* start, bool low_power)
+    {
+        std::thread::id id = std::this_thread::get_id();
+
+        while (!(*start))
+            ;
```
Algo-ryth-mix: why is this `;` on a new line (also why is this not yielding ?)

args/rendering/systems/renderer.hpp
```diff
@@ -171,60 +197,63 @@ namespace args::rendering
...
 
-            if (!shdr)
-                if (!initData(window))
-                    return;
-
             glClearColor(0.3f, 0.5f, 1.0f, 1.0f);
             glClearDepth(0.0f);
             glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
-            glDisable(GL_CULL_FACE);
+            //glDisable(GL_CULL_FACE);
```
Algo-ryth-mix: ah yes ... of course

Algo-ryth-mix (approved):
This code is borderline unreadable:
1. do not change glm mid commits intermingled with other stuff, it is impossible to find the files I am actually interested in
2. add some comments for crying out loud, I am not even talking about doxygen comments, just comments in general, I have no idea what half your changes do 
3. Please stop writing such confusing code just for the hell of it, introduce new variables when the line becomes `batches.dense()[i].dense()[j]`
4. Please try to keep commits under 100 Files, my browser is struggling to even display this many changes

This is approved under the assumption that the next thing you do is documenting what half of this actually does!

applications/test_game/systems/testsystem.hpp
GlynLeine: whoops accidental typo, it works tho

args/core/scheduling/scheduler.cpp
GlynLeine: forgot to add yield here will do so before mergs

args/rendering/systems/renderer.hpp
GlynLeine: I'll remove the line

applications/test_game/systems/testsystem.hpp
GlynLeine: I'll cahnge it before merge

Algo-ryth-mix: yes I made the api, specifically look out for that error.
It is a "hidden" feature xD 
https://github.com/Args-Engine/Args-Engine/blob/main/args/core/filesystem/navigator.cpp
```cpp
 	//find root domain
    // the syntax for inline root-domains has to be <root-domain>:[/\\]+<rest-of-the-path>
    const auto rdIndex = m_path.find_first_of(':');
    if(rdIndex == std::string::npos) return Err(args_fs_error("invalid syntax for path string, no domain delimiter"));

    	root_domain = m_path.substr(0,rdIndex);

        //find the rest of the path ... technically this will resolve //////////lol/test.bin to lol/test.bin just fine,
		//but hopefully no one is actually insane enough to mess with such paths
		//
		//if he/she does, congrats you found a ... "feature"
        const auto pIndex = m_path.find_first_not_of("/\\",rdIndex+1);
        if(pIndex == std::string::npos) return Err(args_fs_error("invalid sytax for path string, last domain delimiter not found"));
          
        to_process = m_p
```

args/core/scheduling/scheduler.cpp
```diff
@@ -6,22 +6,90 @@ namespace args::core::scheduling
...
+        {
+            Scheduler::runnable instruction{};
+
+            {
+                async::readonly_guard guard(m_commandLocks[id]);
+                if (!m_commands[id].empty())
+                {
+                    instruction = m_commands[id].front();
+                    async::readwrite_guard rwguard(m_commandLocks[id]);
+                    m_commands[id].pop();
```
Algo-ryth-mix: isn't this a potential bug ?
consider this scenario:
A & B both lock for reading
A gets front (aka has a command)
B gets front (aka has the same command)

_now_ A locks for deleting with pop()
**B now either fails deleting or can never aquire the lock**

Scenario 1 (B fails deleting) 
 - B triggers an exception (depending on how pop works) or
 - B & A execute the instruction twice

Scenario 2 (B can not aquire the lock)
 - B effectively halts.

args/core/scheduling/scheduler.cpp
```diff
@@ -6,22 +6,90 @@ namespace args::core::scheduling
...
+            instruction = {};
+
+            {
+                async::readonly_guard guard(m_jobQueueLock);
+                if (!m_jobs.empty())
+                {
+                    log::debug("Starting work on a job.");
+                    instruction = m_jobs.front();
+                    async::readwrite_guard rwguard(m_jobQueueLock);
+                    m_jobs.pop();
```
Algo-ryth-mix: same scenario as above

args/core/scheduling/scheduler.cpp
Algo-ryth-mix: potential work around is to check if the job still exists after aquiring the read-lock

args/core/scheduling/scheduler.cpp
```diff
@@ -6,22 +6,90 @@ namespace args::core::scheduling
...
     {
-        for (auto& processChain : m_processChains)
+        for (auto [_, processChain] : m_processChains)
             processChain.exit();
 
-        for (auto& thread : m_threads)
+        m_threadsShouldTerminate = true;
+
+        for (auto [_, thread] : m_threads)
             if (thread->joinable())
```
Algo-ryth-mix: is there a scenario in which our thread are not joinable ?

args/core/scheduling/scheduler.hpp
```diff
@@ -12,208 +13,272 @@
...
+         */
+        void run();
+
+        /**@brief Create a new thread.
+         * @param function Function to run on the thread.
+         * @param ...args Arguments to pass to the function.
+         * @return std::thread::id Id of new thread if thread was created, 0 id if there are no available threads.
+         */
+        template<typename Function, typename... Args >
+        std::thread::id createThread(Function&& function, Args&&... args)
```
Algo-ryth-mix, line 118: consider making this return `common::result<std::thread::id,void>` or introducing an alias `invalid_thread_id` for better readability of error states `std::thread::id()` does not seem like it indicates graceful-failure

Algo-ryth-mix (changes requested):

Algo-ryth-mix (approved):

PR #123, opened 2020-09-21
simple debug line renderer

GlynLeine:
### example:

    #include <rendering/debugrendering.hpp>
    #include <core/core.hpp>

    void foo()
    {
        //                      start,                end,              color,      width, ignore depth test
        debug::drawLine(math::vec3(1, 0, 0), math::vec3(0, 0, 1), math::colors::green, 10, true);
    }


Algo-ryth-mix (approved):
debugRenderPass could use a bit more commenting, other than that quite good

PR #125, opened 2020-09-23
Compute Shaders

Algo-ryth-mix:
# OpenCL
Add's support for OpenCL compute shaders; can be loaded independentily from Rendering

Kernel code is written in a modified version of OpenCL
```c
__kernel void vector_add(__global const int* lhs, __global const int * rhs, __global int * res)
{
	int index = global_get_id(0);
	res[index] = lhs[index] + rhs[index];
}
```


args/core/compute/buffer.hpp
```diff
@@ -0,0 +1,82 @@
...
+        return static_cast<buffer_type>(static_cast<int>(lhs) | static_cast<int>(rhs));
+    }
+
+    /**
+     * @brief Bitwise and operator for type: buffer_type
+     */
+    inline buffer_type& operator|=(buffer_type& lhs,const buffer_type& rhs)
+    {
+        return lhs = lhs | rhs;
+    }
```
Ragingram2: This comment is incorrect.  The operator is a or equals.

args/core/compute/kernel.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+            /**
+             * @brief Determines the "Local Work Size", aka how many Kernels should run in parallel
+             *         This number should not exceed the amount of available cores, a good max is 1024
+             */
+            Kernel& local(size_t);
+
+            /**
+             * @brief Determines the "Global Work Size", aka how many Kernels should run in total
+             *         This number should match your input arrays (but not exceed it!
```
Ragingram2: you missed a ')'

Ragingram2 (changes requested):

Ragingram2 (approved):

args/application/input/inputsystem.hpp
```diff
@@ -217,7 +217,7 @@ namespace args::application
                     }
                     if (inputmap::is_axis(met))
                     {
-                        std::get<0>(m_axes[met][typeHash<Event>()]).clear();
+                        data::m_axes[met][typeHash<Event>()].callback.clear();
```
GlynLeine, line 220: should prolly do that for all of them

args/core/compute/buffer.cpp
```diff
@@ -0,0 +1,85 @@
...
+            m_type = CL_MEM_READ_WRITE;
+            
+
+        cl_int ret;
+
+    	//create buffer
+        m_memory_object = clCreateBuffer(ctx,m_type,m_size,nullptr,&ret);
+        if(ret != NULL)
+        {
+            log::error("clCreateBuffer failed");
```
GlynLeine: since you give buffers a name anyways maybe report the name of the failed buffer

args/application/input/inputsystem.hpp
Algo-ryth-mix: you are correct, will be in next commit

args/core/compute/buffer.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+
+
+
+    Buffer::~Buffer()
+    {
+
+    	//check if this is the last element
+    	//TODO(algo-ryth-mix): make this thread-safe
```
GlynLeine, line 77: using std::atomic for m_ref_count should work. you could then try a compare_exchange_weak in order to really make sure that you are the last one alive.
by using std::atomic the ref count increment will also be thread-safe. 
When using std::atomic don't forget to use the right memory barriers.

args/core/compute/buffer.hpp
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * @class Buffer
+     * @brief Wraps an OpenCL buffer in a more "User-Friendly Interface"
+     *        You would normally obtain these via
+     *        @ref Context::createBuffer()
+     */
+    class ARGS_API Buffer
+    {
+    public:
+        Buffer(cl_context,byte*,size_t,buffer_type,std::string name);
```
GlynLeine: why is only one of these parameters named? XD

args/core/compute/buffer.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+        Buffer(Buffer&& b) noexcept;
+        Buffer(const Buffer& b);
+
+        //copy assign & move assign operator
+        //are deleted since the Buffer
+        //refcounts
+        Buffer& operator=(Buffer&&) noexcept = delete;
```
GlynLeine, line 64: would you not be able to use the same code as in the move and copy constructors instead?

args/core/compute/buffer.hpp
```diff
@@ -0,0 +1,82 @@
...
+        void rename(const std::string& name);
+
+        ~Buffer();
+    private:
+        friend class Program;
+        friend class Kernel;
+
+        std::string m_name;
+        cl_mem m_memory_object;
+        size_t* m_ref_count;
```
GlynLeine: varying use of std::size_t, and size_t instead of args::size_type whilst also using args::byte

args/core/compute/kernel.hpp
```diff
@@ -0,0 +1,183 @@
...
+            void finish() const;
+
+        private:
+
+            buffer_type m_default_mode;
+            std::map<std::string, cl_uint> m_paramsMap;
+            Program* m_prog;
+            cl_kernel m_func;
+            cl_command_queue m_queue;
+            size_t m_global_size;
```
GlynLeine: use of size_t without std::

args/core/compute/kernel.cpp
```diff
@@ -0,0 +1,250 @@
+#include <core/compute/kernel.hpp>
+#include <core/compute/program.hpp>
+#include <core/logging/logging.hpp>
+
+namespace args::core::compute
+{
+    Kernel& Kernel::build_buffer_names()
+    {
+        size_t size;
```
GlynLeine: use of size_t without std::

args/core/compute/kernel.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+            //buffer was read only
+        case CL_MEM_READ_ONLY:
+            //we "write" to a read only buffer because it is readonly for the kernel
+            ret = clEnqueueWriteBuffer(m_queue, buffer.m_memory_object, static_cast<cl_bool>(blocking), 0, buffer.m_size, buffer.m_data, 0, nullptr, nullptr);
+            break;
+
+            //buffer was write only
+        case CL_MEM_WRITE_ONLY:
+            //similarly we read from a buffer that was read only for the kernel
```
GlynLeine: should be: //similarly we read from a buffer that was write-only for the kernel

args/core/compute/kernel.hpp
```diff
@@ -0,0 +1,183 @@
...
+            void param_find(F && func,std::string name)
+            {
+                // lazy build buffer names
+                if (m_paramsMap.empty())
+                {
+                    build_buffer_names();
+                }
+
+                //find and do thing, otherwise print warning
+                if (const auto it = m_paramsMap.find(name); it != m_paramsMap.end())
```
GlynLeine, line 172: if you're using a map anyways, why use std::find? can't you use the accessor operator or .at()?

args/core/defaults/coremodule.hpp
```diff
@@ -5,6 +5,8 @@
 #include <core/filesystem/provider_registry.hpp>
 #include <core/filesystem/basic_resolver.hpp>
 
+#include "core/compute/context.hpp"
```
GlynLeine: please use <> brackets

GlynLeine (approved):

args/core/compute/buffer.hpp
Algo-ryth-mix: potentially, but you would also need to manually decrease the ref count on the previous buffer

args/core/compute/kernel.hpp
Algo-ryth-mix: because I do not want to throw an exception if the element is not in the map

args/core/compute/buffer.cpp
Algo-ryth-mix: it is currently not super important that these are thread-safe(they only represent wrappers around cl_mem), this might come in a future release

PR #130, opened 2020-09-28
Args Audio

Jelled1st:
Audio module for the args engine


args/audio/systems/audiosystem.hpp
```diff
@@ -0,0 +1,32 @@
...
+    public:
+        virtual void setup()
+        {
+            createProcess<&AudioSystem::update>("Audio");
+            bindToEvent<events::component_creation<position>, &AudioSystem::onComponentCreate>();
+        }
+
+        void onComponentCreate(events::component_creation<position>* event)
+        {
+            auto handle = event->entity.get_component_handle<position>();
```
Jelled1st, line 19: should this really be a position?

args/audio/components/audio_source.hpp
```diff
@@ -0,0 +1,70 @@
...
+        * @brief Function to clear the changes that will be applied
+        */
+        void clearChanges()
+        {
+            m_changes ^= m_changes; // Reset
+        }
+
+        ALuint m_sourceId;
+        ALuint m_audioBufferId;
+        audio_segment_handle m_audio_handle;
```
GlynLeine: Buffer should probably be part of the segment

args/audio/data/audio_segment.hpp
```diff
@@ -0,0 +1,45 @@
+#pragma once
+#include <audio/detail/engine_include.hpp>
+
+namespace args::audio
+{
+	struct audio_segment
+	{
+	public:
+        int16* buffer;
```
GlynLeine: rename this to data, and add the OpenAL buffer in here

args/audio/detail/engine_include.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#ifndef ARGS_IMPORT
+    #define ARGS_IMPORT
+    #include <core/core.hpp>
+    #include <core/platform/args_library.hpp>
+#else
+    #include <core/core.hpp>
+#endif // !ARGS_IMPORT
```
GlynLeine: This is no longer needed, You can just straight up include <core/core.hpp> instead

args/audio/systems/audiosystem.hpp
```diff
@@ -0,0 +1,93 @@
...
+ * @file audiosystem.hpp
+ */
+
+inline void openal_error()
+{
+    ALCenum error;
+    error = alGetError();
+    if (error != AL_NO_ERROR)
+        args::log::warn("ERROR: OpenAl error: {}", error);
+}
```
GlynLeine: You shouldn't leave global functions in the global namespace...

args/audio/systems/audiosystem.hpp
```diff
@@ -0,0 +1,93 @@
...
+        position m_listenerPosition;
+        std::unordered_map <ecs::component_handle<audio_source>, position > m_sourcePositions;
+
+        struct ARGS_API data
+        {
+            static ALCdevice* alDevice;
+            static ALCcontext* alContext;
+            static unsigned int sourceCount;
+            static unsigned int listenerCount;
+        };
```
GlynLeine: this is no longer needed, statics can just straight up be part of AudioSystem without needing ARGS_API. ARGS_API is deprecated and unnecessary.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+		const ALchar* ALCExtensions = alcGetString(data::alDevice, ALC_EXTENSIONS);
+		ALCint srate;
+		alcGetIntegerv(data::alDevice, ALC_FREQUENCY, 1, &srate);
+		log::info("OpenAL info:\n\n\t\t\tOpenAL information\n\tVendor: {}\n\tVersion: {}\n\tRenderer: {}\n\tOpenAl Extensions: {}\n\tALC Extensions: {}\n\tDevice samplerate: {}\n",
+			vendor, version, renderer, openALExtensions, ALCExtensions, srate);
+	}
+
+	inline void AudioSystem::update(time::span deltatime)
+	{
+		alcMakeContextCurrent(data::alContext);
```
GlynLeine: what if making the context current fails because another thread is utilizing the context?

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+			alListener3f(AL_VELOCITY, vel.x, vel.y, vel.z);
+		}
+
+		alcMakeContextCurrent(nullptr);
+	}
+
+#pragma region Component creation&destruction
+	inline void AudioSystem::onAudioSourceComponentCreate(events::component_creation<audio_source>* event)
+	{
+		alcMakeContextCurrent(data::alContext);
```
GlynLeine: what if making the context current fails because another thread is utilizing the context?

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+		handle.write(a);
+		++data::sourceCount;
+
+		alcMakeContextCurrent(nullptr);
+	}
+
+
+	inline void AudioSystem::onAudioSourceComponentDestroy(events::component_destruction<audio_source>* event)
+	{
+		alcMakeContextCurrent(data::alContext);
```
GlynLeine: what if making the context current fails because another thread is utilizing the context?

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+
+		handle.write(a);
+		--data::sourceCount;
+
+		alcMakeContextCurrent(nullptr);
+	}
+
+	inline void AudioSystem::onAudioListenerComponentCreate(events::component_creation<audio_listener>* event)
+	{
+		alcMakeContextCurrent(data::alContext);
```
GlynLeine: what if making the context current fails because another thread is utilizing the context?

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+			setListener(event->entity.read_component<position>(), event->entity.read_component<rotation>());
+			m_listenerPosition = event->entity.read_component<position>();
+		}
+		handle.write(a);
+		alcMakeContextCurrent(nullptr);
+	}
+
+	inline void AudioSystem::onAudioListenerComponentDestroy(events::component_destruction<audio_listener>* event)
+	{
+		async::readwrite_guard guard(*m_lock);
```
GlynLeine: Here you use the context lock, but not in the other places. Also, m_lock isn't a very descriptive name.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+		}
+		
+		alSourcei(source.m_sourceId, AL_BUFFER, source.m_audioBufferId);
+
+		return true;
+	}
+
+	inline void AudioSystem::setDistanceModel(ALenum distanceModel)
+	{
+		alDistanceModel(distanceModel);
```
GlynLeine, line 241: does the context not need to be current? if so please add a comment.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	inline void AudioSystem::setListener(position p, rotation r)
+	{
+		// Position - invert x for left-right hand coord system conversion
```
GlynLeine, line 249: does the context not need to be current? if so please add a comment.

GlynLeine:
please merge develop into this branch first in order to resolve the conflicts

args/audio/components/audio_source.hpp
```diff
@@ -0,0 +1,69 @@
...
+        * @brief Function to get the current pitch
+        */
+        float getPitch() const { return m_pitch; }
+        /**
+        * @brief Function to set the gain for the audio source
+        * @param const float gain: new gain value
+        */
+        void setGain(const float gain)
+        {
+            m_changes |= 2; // set b1
```
GlynLeine: a more readable way to do this is to create an enum with pitch = 1 and gain = 2 and then set m_changes |= changes::gain

args/audio/data/audio_segment.cpp
```diff
@@ -0,0 +1,48 @@
...
+            if (m_segments.count(id))
+                return { id };
+        }
+
+        // Segment is loaded for the first time
+        auto result = fs::AssetImporter::tryLoad<audio_segment>(file, settings);
+        if (result != common::valid)
+        {
+            log::error("Audio file wrong!");
+            //log::error("Error while loading file: {}, {}", file.get_filename(), result.get_error().what());
```
GlynLeine: I think we can remove the first log and uncomment the second log again

args/audio/data/importers/audio_importers.hpp
```diff
@@ -0,0 +1,14 @@
...
+#include <audio/data/audio_segment.hpp>
+
+namespace args::audio
+{
+    struct mp3_audio_loader : public fs::resource_converter<audio_segment, audio_import_settings>
+    {
+        virtual common::result_decay_more<audio_segment, fs_error> load(const fs::basic_resource& resource, audio_import_settings&& settings) override;
+        static std::shared_ptr<ALCcontext> context;
+    private:
+        static async::readonly_rw_spinlock contextLock;
```
GlynLeine: this is not the same lock as the one in audiosystem.hpp, audiosystem and the audio loader can thus both lock different locks at the same time. meaning they can both try to use the context at the same time.

args/audio/systems/audiosystem.hpp
```diff
@@ -0,0 +1,90 @@
...
+        */
+        void onAudioListenerComponentDestroy(events::component_destruction<audio_listener>* event);
+
+        void update(time::span deltatime);
+
+        static void setDistanceModel(ALenum distanceModel);
+
+    private:
+        bool initSource(audio_source& source);
+        async::readonly_rw_spinlock m_contextLock;
```
GlynLeine: this lock should be the only lock used to lock the context, other classes that want to use the context also need to lock this exact lock.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+		alDevice = alcOpenDevice(NULL);
+		if (!alDevice)
+		{
+			// Failed to create alcDevice
+			log::error("OpenAl device failed to open");
+#if defined(AUDIO_EXIT_ON_FAIL)
+			raiseEvent<events::exit>();
+#endif
+			return;
+		}
```
GlynLeine: In setup everything is still single threaded so locking the lock isn't necessary yet.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+	{
+		const ALchar* vendor = alGetString(AL_VENDOR);
+		const ALchar* version = alGetString(AL_VERSION);
+		const ALchar* renderer = alGetString(AL_RENDERER);
+		const ALchar* openALExtensions = alGetString(AL_EXTENSIONS);
+		const ALchar* ALCExtensions = alcGetString(alDevice, ALC_EXTENSIONS);
+		ALCint srate;
+		alcGetIntegerv(alDevice, ALC_FREQUENCY, 1, &srate);
+		log::info("OpenAL info:\n\n\t\t\tOpenAL information\n\tVendor: {}\n\tVersion: {}\n\tRenderer: {}\n\tOpenAl Extensions: {}\n\tALC Extensions: {}\n\tDevice samplerate: {}\n",
+			vendor, version, renderer, openALExtensions, ALCExtensions, srate);
```
GlynLeine, line 80: does the context need to be current for this or is it only called from places where the context is already current? if so please add a comment.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+			if (source.m_changes & 1)
+			{
+				// Pitch has changed
+				alSourcef(source.m_sourceId, AL_PITCH, source.getPitch());
+			}
+			if (source.m_changes & 2)
+			{
+				// Gain has changed
+				alSourcef(source.m_sourceId, AL_GAIN, source.getGain());
+			}
```
GlynLeine: same thing as before maybe think of using an enum to make this more readable and maintainable

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+
+		{
+			async::readwrite_guard guard(m_contextLock);
+			alcMakeContextCurrent(alContext);
+			// NOTE TO SELF:
+			// REMOVE THE AUTO PLAY (TESTING PURPOSE)
+			log::debug("playing sound");
+			alSourcePlay(a.m_sourceId);
+			alcMakeContextCurrent(nullptr);
+		}
```
GlynLeine: technically this could also be done from the test system.

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+
+
+	inline void AudioSystem::onAudioSourceComponentDestroy(events::component_destruction<audio_source>* event)
+	{
+		auto handle = event->entity.get_component_handle<audio_source>();
+		audio_source a = handle.read();
+
+		m_sourcePositions.erase(handle);
+
+		handle.write(a);
```
GlynLeine: what's the point of the read and the write here?

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+			event->entity.remove_component<audio_listener>();
+		}
+		else
+		{
+			// listener count == 1
+			m_listenerEnt = event->entity;
+			setListener(event->entity.read_component<position>(), event->entity.read_component<rotation>());
+			m_listenerPosition = event->entity.read_component<position>();
+		}
+		handle.write(a);
```
GlynLeine: what's the point of the read and write here?

args/audio/systems/audiosystem.inl
```diff
@@ -0,0 +1,273 @@
...
+		// Moved alBufferData to audio_importers.cpp
+		/*{
+			async::readonly_guard guard(lock);
+			alBufferData(source.m_audioBufferId, AL_FORMAT_MONO16, segment.buffer, segment.samples * sizeof(int16), segment.sampleRate);
+		}*/
+		
+		alSourcei(source.m_sourceId, AL_BUFFER, segment.audioBufferId);
+		alcMakeContextCurrent(nullptr);
+
+		return true;
```
GlynLeine: this can never fail? if so maybe remove the return bool and return void instead. right now this function seems very scary. it tells me it's failable, and if it fails the engine might just straight up quit... so every time I add an audio source there's a chance the engine will just fail and quit.

args/audio/components/audio_source.hpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    struct audio_source
+    {
+        friend class AudioSystem;
+    public:
+        enum sound_properties
+        {
+            pitch = 1,
+            gain = 2,
+        };
```
GlynLeine, line 16: if this is used more than once in the audio library, maybe make it a global enum in the args::audio namespace instead of a member of the audio_source class.

args/audio/systems/audiosystem.hpp
```diff
@@ -0,0 +1,90 @@
...
+        */
+        void queryInformation();
+        void setListener(position p, rotation r);
+        ecs::entity_handle m_listenerEnt;
+
+        position m_listenerPosition;
+        std::unordered_map <ecs::component_handle<audio_source>, position > m_sourcePositions;
+
+        static ALCdevice* alDevice;
+        static ALCcontext* alContext;
```
GlynLeine: the pointer in audio_importer is a shared pointer but this one isn't? also aren't OpenAL pointers supposed to be cleaned up by using an OpenAL function like alcDestroyContext() or something like that? instead of being deleted using the C++ delete operator?

args/audio/data/importers/audio_importers.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+#include <core/core.hpp>
+#include <audio/data/audio_segment.hpp>
+
+namespace args::audio
+{
+    struct mp3_audio_loader : public fs::resource_converter<audio_segment, audio_import_settings>
+    {
+        virtual common::result_decay_more<audio_segment, fs_error> load(const fs::basic_resource& resource, audio_import_settings&& settings) override;
+        static std::shared_ptr<ALCcontext> context;
```
GlynLeine: as mentioned below, this should probably not be deleted by shared pointers but instead by alcDestroyContext().

GlynLeine (approved):

kukash (approved):

PR #133, opened 2020-09-29
Physics manifold filling

Developer-The-Great:


args/physics/PhysicsStatics.h
```diff
@@ -50,12 +50,12 @@ namespace args::physics
...
         {
             float currentMaximumSeperation = std::numeric_limits<float>::lowest();
- 
 
-            for (const auto face : convexB->GetHalfEdgeFaces())
+
+            for ( auto face : convexB->GetHalfEdgeFaces())
             {
                 //get inverse normal
                 math::vec3 seperatingAxis = transformB * math::vec4( face->normal, 0);
```
GlynLeine, line 61: this file is a .h but contains c++ code. it should be a .hpp file

Ragingram2 (approved):

args/physics/colliders/convexcollider.cpp
```diff
@@ -1,49 +1,96 @@
...
 {
     void ConvexCollider::CheckCollisionWith(ConvexCollider* convexCollider, physics_manifold& manifold) 
     {
-      
 
+        //--------------------- Check for a collision by going through the edges and faces of both polyhedrons  --------------//
         //'this' is colliderA and 'convexCollider' is colliderB
 
-        HalfEdgeFace* ARefFace = nullptr;
+        HalfEdgeFace** ARefFace = &convexCollider->GetHalfEdgeFaces().at(0);
```
GlynLeine, line 15: is this actually supposed to be pointing to the first item in the container, or just to the data that's in the container? otherwise maybe use a .data() function or something like that.

GlynLeine:
also please resolve the merge conflicts

args/physics/PhysicsStatics.h
```diff
@@ -50,12 +50,12 @@ namespace args::physics
...
         static bool FindSeperatingAxisByExtremePointProjection(ConvexCollider* convexA
-            , ConvexCollider* convexB, const math::mat4& transformA, const math::mat4& transformB, HalfEdgeFace* refFace, float& maximumSeperation) 
+            , ConvexCollider* convexB, const math::mat4& transformA, const math::mat4& transformB, HalfEdgeFace** refFace, float& maximumSeperation) 
         {
             float currentMaximumSeperation = std::numeric_limits<float>::lowest();
- 
 
-            for (const auto face : convexB->GetHalfEdgeFaces())
+
+            for ( auto face : convexB->GetHalfEdgeFaces())
```
Algo-ryth-mix, line 58: this file should either be `physicsstatics.hpp` or `physics_statics.hpp` (I prefer the later for readability)

Algo-ryth-mix (changes requested):

args/physics/colliders/convexcollider.cpp
Developer-The-Great: it doesnt actually matter which item it initially points to. I just needed to initialize it with something because in the function below it its treated as an out parameter.

Algo-ryth-mix (approved):

PR #153, opened 2020-10-08
image in core and window system overhoal

GlynLeine:


GlynLeine:
Had to fix a missing null check for an audio segment in order to run it. I also cleaned up the OpenAL context info logging.

GlynLeine:
image

Algo-ryth-mix (approved):

args/core/data/image.cpp
```diff
@@ -0,0 +1,339 @@
...
+            }
+        }
+
+        {
+            async::readwrite_guard guard(m_colorsLock);
+            m_colors.emplace(std::make_pair(id, std::unique_ptr<std::vector<math::color>>(ptr)));
+        }
+
+        return output;
+    }
```
Developer-The-Great, line 241: this is quite a huge function, some comments might be necessary

Developer-The-Great (changes requested):

Developer-The-Great (approved):

PR #156, opened 2020-10-13
Feature/debug renderer double buffer

GlynLeine:


Developer-The-Great (approved):

args/core/math/glm/detail/type_vec3.hpp
```diff
@@ -22,6 +22,13 @@ namespace args::core::math
 		typedef vec<3, T, Q> type;
 		typedef vec<3, bool, Q> bool_type;
 
+        static const type up;
+        static const type down;
+        static const type right;
+        static const type left;
+        static const type forward;
+        static const type backward;
+
```
Algo-ryth-mix, line 31: what is the use of these ? as far as I can tell these should be initialized here!

args/core/math/glm/detail/type_vec3.inl
```diff
@@ -165,26 +178,26 @@ namespace args::core::math
...
-		return y;
-			case 2:
-		return z;
+		case 0:
+			return x;
+		case 1:
+			return y;
+		case 2:
+			return z;
 		}
```
Algo-ryth-mix, line 193: you could also add an operator for char types to allow for this syntax: `v['x']` which would be useful if we want to expose this operator to a scripting language

args/core/math/glm/detail/type_vec3.hpp
GlynLeine: since they're static they can't be initiliased here

Algo-ryth-mix: they are also const

args/core/math/glm/detail/type_vec3.inl
GlynLeine: would be usefull but this is just default glm

args/core/math/glm/detail/type_vec3.hpp
Algo-ryth-mix: ah well they aren't pod so I guess they can't be static constexpr, which would probably be the ideal case

PR #175, opened 2020-10-26
[Add] ShellInvoke for windows and linux

GlynLeine:


legion/engine/core/platform/shellinvoke.hpp
```diff
@@ -0,0 +1,68 @@
...
+            file = command.substr(0, seperator);
+            params = command.substr(seperator);
+        }
+
+        paramsvec.assign(params.begin(), params.end());
+        common::replace_items(paramsvec, " ", "\0");
+        paramsvec.push_back('\0');
+
+        return execvp(file.c_str(), paramsvec.data()) != -1;
+    }
```
Algo-ryth-mix: there is a few issues here, but it boils down to this:
execvp, (much like what main would get) expects a `const char **`(aka an array of null terminated strings) with a NULL pointer as the last field, not a const char * with null terminations

Algo-ryth-mix (changes requested):

legion/engine/core/platform/shellinvoke.hpp
Algo-ryth-mix: Another small note, the first parameter passed to file via `execvp(<file>,<parameters>)` must be the name of the command itself otherwise it is not going to receive anything.

Algo-ryth-mix: Here is a minimal verified (Ubuntu 20.04 LTS) example, that is slighlty leaky (no matching `delete` for the `new`s in the tokenization step) but works fine:
```cpp
#include <unistd.h>
#include <string>
#include <sstream>
#include <vector>
#include <algorithm>
#include <cstring>

inline std::vector<char*> tokenize(const std::string& source)
{
    std::vector<char*> result;

    auto iss = std::stringstream(source);
    auto str = std::string();

    while (iss >> str)
    {
        const char * r = str.c_str();
        char * x = new char[str.size()];
        memcpy(x,r,str.size());
        result.push_back(x);
    } 
    return result;
}

inline bool ShellInvoke(const std::string& command)
{
    std::string file;
    std::vector<char*> paramsvec;

    auto seperator = command.find_first_of(' ');
    if (seperator == std::string::npos)
    {
        file = command;
    }
    else
    {
        file = command.substr(0, seperator);    
    }

    paramsvec = tokenize(command);
    paramsvec.push_back(NULL);

    return execvp(file.c_str(), paramsvec.data()) != -1;
}

int main(){

    ShellInvoke("touch test.txt");

}
```

Developer-The-Great (approved):

Algo-ryth-mix (approved):

PR #180, opened 2020-10-28
Rigidbody-Friction

Developer-The-Great:
- Constraint Based Friction Implemented
- Fixed quaternion integration bug
- Created Friction and Stacking Unit tests

### DIFF
--- a/legion/engine/physics/physics_contact.hpp
+++ b/legion/engine/physics/physics_contact.hpp
@@ -2,6 +2,7 @@
 
 #include <physics/components/rigidbody.hpp>
 #include <core/core.hpp>
+#include <physics/data/identifier.hpp>
 
 namespace legion::physics
 {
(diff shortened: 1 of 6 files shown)


Algo-ryth-mix (approved):

legion/engine/physics/physics_contact.hpp
```diff
@@ -121,73 +154,178 @@ namespace legion::physics
...
+
+        float calculateEffectiveMassOnNormal(math::vec3& normal)
+        {
+            math::vec3 Ra, Rb, minRaCrossN, RbCrossN;
+            calculateJacobianComponents(normal, Ra, minRaCrossN, Rb, RbCrossN);
+
+            auto rbRef = rbRefHandle.read();
+            auto incRef = rbIncHandle.read();
+
+            float maUnit = rbRefHandle ? rbRef.inverseMass : 0.0f;
```
Jelled1st, line 299: Shouldn't the rbRefHandle be tested before reading?

Jelled1st (approved):

legion/engine/physics/physics_contact.hpp
Developer-The-Great: iirc, reading a null component handle gives the default values of the component. But that does mean that checking 
rbRefHandle? is actually not necessary. I'll fix it in my next pull request. Thanks ^^

PR #197, opened 2020-11-18
seperate camera logic to simple camera controller

GlynLeine:
Also had to remove some Physics and OpenCL stuff because testsystem wouldn't compile.
This is also why I asked you two, @Algo-ryth-mix and @Developer-The-Great, because you might need to add it back in


Developer-The-Great (approved):

Algo-ryth-mix:
I'll check it out tomorrow, but OpenCL should A not break anything and B not be broken by being removed. If it causes compiler errors now however I need to take a look

PR #203, opened 2020-11-20
Feature/scene management

Ragingram2:
I am going to finish up the scene manager and make it easier to use


Algo-ryth-mix:
Make sure to change the base branch once changes are present otherwise the CI pipeline will break.
Also we ain't mergin straight into main around here

PR #276, opened 2021-01-08
[LAP-9] bloom postprocessing effect

tuhri444:
Bloom post-processing effect, simple as that. Some minor changes were made to stdio.shinc  and because of that all the other shaders, but those are neglectable.
image


Algo-ryth-mix (changes requested):
Something weird with LFS is going on here, I'll check it out and make sure that it isn't an issue before approving, other than that looks good

kukash (approved):

Algo-ryth-mix (approved):

PR #289, opened 2021-01-20
fix docs action errors

GlynLeine:


Algo-ryth-mix (approved):
this works but maybe a macro like this
```cpp
#if defined DOXY_EXCLUDE
CDOXY(X)
#else
CDOXY(X) ,X
#endif
```

could aid in cleaning this up a bit 


for example
```cpp
#if !defined(DOXY_EXCLUDE)
    template <class T, class...Any,
        std::enable_if_t<!(std::is_base_of_v<result_ident, T> && sizeof...(Any) == 0), int> = 0>
#else
    template <class T, class...Any>
#endif
```
Would be 
```cpp
template <class T, class ... Any 
CDOXY(std::enable_if_t<!(std::is_base_of_v<result_ident, T> && sizeof...(Any) == 0),int> = 0)>
```

GlynLeine:
> this works but maybe a macro like this
> 
> ```c++
> #if defined DOXY_EXCLUDE
> CDOXY(X)
> #else
> CDOXY(X) ,X
> #endif
> ```
> 
> could aid in cleaning this up a bit
> 
> for example
> 
> ```c++
> #if !defined(DOXY_EXCLUDE)
>     template <class T, class...Any,
>         std::enable_if_t<!(std::is_base_of_v<result_ident, T> && sizeof...(Any) == 0), int> = 0>
> #else
>     template <class T, class...Any>
> #endif
> ```
> 
> Would be
> 
> ```c++
> template <class T, class ... Any 
> CDOXY(std::enable_if_t<!(std::is_base_of_v<result_ident, T> && sizeof...(Any) == 0),int> = 0)>
> ```

what does the C in CDOXY stand for in this case?

Algo-ryth-mix:
Comma

Ragingram2 (approved):

PR #294, opened 2021-01-24
Feature/material loading from models

GlynLeine:
Add support for loading materials straight from the model file for both OBJ and GLB and GLTF
**blender:**
image
**legion:**
image


kukash (approved):

legion/engine/core/filesystem/view.hpp
```diff
@@ -7,6 +7,9 @@
 
 #include <Optick/optick.h>
 
+#include <string_view>
```
Algo-ryth-mix: using namespace in header files breaks things

Algo-ryth-mix (changes requested):

Algo-ryth-mix (approved):

PR #325, opened 2021-04-29
Use context folder when loading .mtl textures

LeonBrands:
The model loader for .obj files uses the .mtl file to load in textures but didn't use the import settings' context folder to localize


GlynLeine:
have you tested whether this works?

LeonBrands:
Just did, it works

GlynLeine (approved):

PR #341, opened 2021-06-08
Subfeature/angle based threshold for edge collisions

Developer-The-Great:


GlynLeine (approved):
Build action states that `HalfEdgeFace::to_string` doesn't always return a value. It probably also should be a switch statement, and not a series of if statements.
it also warns that `BroadphaseUniformGrid::reConstruct` returns an adress to deallocated stack memory.

Algo-ryth-mix (approved):
There seem to be a lot of double comments
I.e:
```
///*
// *
// */
```

PR #344, opened 2021-07-20
Subfeature/fixing adjacent face merging

Developer-The-Great:
face merging has been reworked, this feature would make quickhull "feature complete".


LeonBrands:
I'm sure I'm missing a lot of context but it seems like some of the file and method names don't follow the legion naming conventions;
* A file name cannot have any uppercase letters
* Methods of non "stl-likes" should use camelCase naming and methods of "stl-likes" should use stl-like snake_case naming.

GlynLeine (approved):
Honestly besides the amount of raw pointers that are being used and the code convention breaks idrk what else to comment. and both of those require major refactoring throughout the entire physics engine at some point. the pointer use is also a design decision that we maybe should talk about together sometime.

PR #354, opened 2021-08-05
Allow for other repositories to make docs

Algo-ryth-mix:
Currently the docs action would completely purge the remote server when it rebuilds, this is nice for a sane docs environment, but a little bit bad if other projects also want to write to that directory, this pr replaces the hard purge `rm -rf ~\public_html\docs` with an approach that preserves symlinks `find ~\public_html\docs ! -type l -delete`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -41,7 +41,7 @@ jobs:
       uses: appleboy/ssh-action@master
       with:
         script: |
-         rm -rf ~/public_html/docs
+         find ~/public_html/docs ! -type l -delete || true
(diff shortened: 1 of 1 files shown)


LeonBrands (approved):

GlynLeine (commented):
do we wanna merge this into main? or into some other branch?

LeonBrands:
I imagine docs are only built on main thus this should be pushed to main?

GlynLeine:
> I imagine docs are only built on main thus this should be pushed to main?

you are completely right

GlynLeine (approved):

PR #357, opened 2021-08-22
Subfeature/asset pipeline impl

GlynLeine:


legion/engine/core/assets/functionalbinds.hpp
```diff
@@ -4,55 +4,63 @@
 namespace legion::core::assets
 {
     template<typename AssetType>
```
Ragingram2, line 6: Why are these no longer static?

Ragingram2 (approved):

legion/engine/core/assets/functionalbinds.hpp
GlynLeine: no idea tbh, they should be static... I'll change it back with the next PR

PR #359, opened 2021-08-30
Subfeature/new mesh asset

GlynLeine:


legion/engine/core/assets/assetloader.hpp
```diff
@@ -19,19 +19,16 @@ namespace legion::core::assets
         using import_cfg = import_settings<AssetType>;
         using progress_type = async::async_progress<common::result<asset_ptr>>;
 
-    private:
-        id_type m_loaderId;
-
     protected:
         template<typename... Args>
-        asset_ptr create(id_type nameHash, const std::string& name, Args&&... args) const;
+        asset_ptr create(id_type nameHash, Args&&... args) const;
```
Ragingram2, line 24: any reason why we don't want to give names to assets when creating them?

GlynLeine: because the loader doesn't need to know about it anymore, the assetcache handles it already

Ragingram2 (approved):

PR #362, opened 2021-09-02
Feature/asset management

### LINKED ISSUE #330: Centralized Asset Management
Right now there is no reference implementation for asset management. So each new asset needs to create its own systems and API. this also causes the API around assets to not be coherent. Additionally adding new custom assets is very difficult.

GlynLeine:
Requires PR #361


applications/sandbox/source.cpp
```diff
@@ -1,20 +1,20 @@
...
 {
     using namespace legion;
-
-    engine->reportModule<app::ApplicationModule>();
-    engine->reportModule<gfx::RenderingModule>();
-    engine->reportModule<audio::AudioModule>();
+    log::filter(log::severity_debug);
+    //engine->reportModule<app::ApplicationModule>();
+    //engine->reportModule<gfx::RenderingModule>();
+    //engine->reportModule<audio::AudioModule>();
```
Ragingram2: we probably want to leave these uncommented for develop.(At least app and rendering)

Ragingram2 (dismissed):

Ragingram2 (approved):

PR #366, opened 2021-09-28
Hotfix/gltf loading

GlynLeine:


legion/engine/rendering/pipeline/default/stages/meshrenderstage.cpp
```diff
@@ -95,6 +95,98 @@ namespace legion::rendering
 
         for (auto [material, instancesPerMaterial] : *batches)
         {
+            if (material.get_name() == "Test")
```
Ragingram2, line 98: is this all still neccesary? Is Leon using the Test material only?

Ragingram2 (approved):

PR #376, opened 2021-11-25
switched filemanip read_file() to std::ifstream

LeonBrands:
Prevents access violations on read-only files


LeonBrands:
the rest of filemanip.hpp should probably be updated to modern C++ too but I don't have the time to do so right now

GlynLeine (approved):

PR #391, opened 2022-01-31
Subfeature/physics compiles on new core

Developer-The-Great:
- updates the bare minimum necessary to make physics compile on new core
- removes dynamic destruction implementation from the codebase


applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -126,73 +120,58 @@ namespace legion::physics
         //stabilityComparisonScene();
```
LeonBrands, line 120: give physics_test_system a quick manual re-review to check for formatting and commented out code. clean up where possible, not gonna review it too harshly because it's a sandbox class

legion/engine/physics/module/physicsmodule.hpp
```diff
@@ -0,0 +1,29 @@
...
+            registerComponentType<identifier>();
+        }
+
+        virtual priority_type priority() override
+        {
+            return 20;
+        }
+    };
+}
+
```
LeonBrands, line 28: cleanup excessive newline

LeonBrands (approved):
minor changes that you may follow up on but it's not huge/breaking so lgtm

applications/sandbox/systems/physics_test_system.cpp
Developer-The-Great: some of the added commented code is related to the recursive template bug, would it be okay with you if I addressed this in my next subfeature?

LeonBrands: That's fine!

PR #394, opened 2022-02-02
Task/rework startup and shutdown

GlynLeine:


GlynLeine:
this is in preparation for being able to run multiple instances of the engine per program. useful for both the editor as well as for unit tests

legion/engine/core/ecs/registry.cpp
```diff
@@ -3,6 +3,24 @@
...
+    {
+        auto& families = getFamilies();
+        if (families.count(id)) // Check and fetch in order to avoid a possibly unnecessary allocation and deletion.
+            return families.at(id).get();
+
+        // Allocate and emplace if no item was found.
+        return families.emplace(
+            id,
+            std::move(ptr)
+        ).first->second.get(); // std::pair<iterator, bool>.first --> iterator<std::pair<key, value>>->second --> std::unique_ptr.get() --> component_pool_base*
```
Developer-The-Great, line 16: is this suppose to be a comment or is this commented code?

legion/engine/core/engine/engine.hpp
```diff
@@ -3,7 +3,10 @@
 #include <core/engine/module.hpp>
 #include <core/types/primitives.hpp>
 #include <core/types/meta.hpp>
+#include <core/async/spinlock.hpp>
+#include <core/containers/pointer.hpp>
 #include <core/logging/logging.hpp>
+
 #include <argh.h>
```
Developer-The-Great, line 10: is there a reason why a space was added here?

Developer-The-Great (approved):
not too familiar with the code being changed here, but only a few minor formatting concerns, LGTM

legion/engine/core/ecs/registry.cpp
GlynLeine: this is not commented out code, it's a comment saying which types are involved. It can get a bit confusing with such layered types

legion/engine/core/engine/engine.hpp
GlynLeine: yeah, to separate the legion includes from external includes

PR #395, opened 2022-02-02
Subfeature/rigidbody dynamics works on new core

Developer-The-Great:
- Ensures rigid body dynamics implementation is functional in new core
- removes old unused quickhull implementation
- Ensures quickhull test scene and stability test scene is functional in new core
- fixed requested physics_test_system formatting changes from #391


legion/engine/physics/systems/physicssystem.cpp
```diff
@@ -21,10 +21,18 @@ namespace legion::physics
...
-        createProcess<&PhysicsSystem::fixedUpdate>("Physics", m_timeStep);
-
         m_broadPhase = std::make_unique<BroadphaseUniformGridNoCaching>(math::vec3(3, 3, 3));
+    }
+
+    void PhysicsSystem::update(legion::time::span deltaTime)
+    {
+        accumulator += deltaTime;
 
+        while (accumulator > m_timeStep)
```
LeonBrands: This could accumulate and cause long freezes if fixedUpdate() takes more than m_timeStep, might be worth to consider some kind of limit where it resets accumulation

Developer-The-Great: I'm going to clamp the number of fixed updates in one frame instead as resetting the accumulator would cause us to "lose" physics updates and I would prefer physics to be deterministic. Clamping will allow the physics engine to "catch up" after a lag spike

LeonBrands: Sounds good

applications/sandbox/imgui.ini
```diff
@@ -39,7 +39,7 @@ Size=450,444
 Collapsed=0
 
 [Window][###]
-Pos=46,71
```
LeonBrands: imgui.ini changes seem accidental, probably shouldn't be pushed (also @GlynLeine maybe imgui.ini should be ignored instead?)

GlynLeine: yeah it should, might as well do it now

GlynLeine:
i wanna have a look in a bit

GlynLeine (commented):
i wanna have a look in a bit

applications/sandbox/sandbox.vcxproj
```diff
@@ -101,7 +101,7 @@
       <OptimizeReferences>true</OptimizeReferences>
       <GenerateDebugInformation>false</GenerateDebugInformation>
       <AdditionalLibraryDirectories>$(SolutionDir)lib\;$(SolutionDir)deps\lib\;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
-      <AdditionalDependencies>legion-core.lib;legion-application.lib;legion-rendering.lib;legion-audio.lib;OpenCL.lib;OpenAL32.lib;glfw3.lib;Voro++.lib;OptickCore.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
GlynLeine, line 104: damn i rly gotta make a comment on a piece of code to stop it from auto merging...

LeonBrands (approved):
lgtm

legion/engine/physics/systems/physicssystem.cpp
```diff
@@ -21,10 +21,22 @@ namespace legion::physics
...
         m_broadPhase = std::make_unique<BroadphaseUniformGridNoCaching>(math::vec3(3, 3, 3));
+    }
+
+    void PhysicsSystem::update(legion::time::span deltaTime)
+    {
+        m_accumulator += deltaTime;
 
+        int currentFrameFixedUpdateCount = 0;
+
+        while (m_accumulator > m_timeStep && currentFrameFixedUpdateCount < m_maxTimeStepPerFrame)
```
GlynLeine: if `currentFrameFixedUpdateCount` (which is a little overly descriptive) is mainly used to compare against `m_maxTimeStepPerFrame` (maybe m_maxIntervals?) they should be of the same type. at the very least of the same signedness. so i'd suggest making `currentFrameFixedUpdateCount` a size_type

legion/engine/physics/systems/physicssystem.hpp
```diff
@@ -33,24 +26,23 @@ namespace legion::physics
...
             std::vector<byte> hasRigidBodies;
 
             {
                 OPTICK_EVENT("Fetching data");
 
-                rigidbodies.reserve(manifoldPrecursorQuery.size());
+                rigidbody emptyRigidbody;
+                std::reference_wrapper<rigidbody> wrapperToEmptyBody = emptyRigidbody;
+
+                rigidbodies.resize(manifoldPrecursorQuery.size(), wrapperToEmptyBody);
```
GlynLeine:
```cpp
rigidbody emptyRigidbody;
rigidbodies.resize(manifoldPrecursorQuery.size(), std::ref(emptyRigidBody));
```

legion/engine/physics/systems/physicssystem.hpp
```diff
@@ -129,9 +121,12 @@ namespace legion::physics
     private:
 
         static std::unique_ptr<BroadPhaseCollisionAlgorithm> m_broadPhase;
-        const float m_timeStep = 0.02f;
 
+        const float m_timeStep = 0.02f;
+        float m_accumulator = 0.0f;
 
+        size_t m_maxTimeStepPerFrame = 3;
```
GlynLeine: `size_t` is part of the `std::` namespace, so this would either require that namespace attached, or `size_type` should be used. `size_type` is the rythe default and technically more correct in this case.

legion/engine/physics/physicsconstants.hpp
```diff
@@ -6,7 +6,7 @@ namespace legion::physics::constants
 {
     static math::vec3 gravity = math::vec3(0, -9.81, 0);
 
-    static constexpr int contactSolverIterationCount = 6;
+    static constexpr int contactSolverIterationCount = 9;
 
     static constexpr int frictionSolverIterationCount = 4;
```
GlynLeine: I think generally things like counts and sizes are better kept as size_type instead of integers, even if not only for semantics, then for the promise that it won't be negative.

GlynLeine (approved):

PR #397, opened 2022-02-09
Feature/port physics module to new core

Developer-The-Great:


applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+        suzzaneH = rendering::ModelCache::create_model("suzanne", "assets://models/suzanne.glb"_view);
+        teapotH = rendering::ModelCache::create_model("sah", "assets://models/gnomecentered.obj"_view);
+
+        #pragma endregion
+
+        #pragma region Input binding
+
+        app::InputSystem::createBinding<extendedPhysicsContinue>(app::inputmap::method::M);
+        app::InputSystem::createBinding<nextPhysicsTimeStepContinue>(app::inputmap::method::N);
+        //app::InputSystem::createBinding<spawnEntity>(app::inputmap::method::MOUSE_LEFT);
```
LeonBrands: remove commented out code

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+            auto sun = createEntity();
+            sun.add_component<rendering::mesh_renderable>(mesh_filter(directionalLightH.get_mesh()), rendering::mesh_renderer(directionalLightMH));
+            sun.add_component<rendering::light>(rendering::light::directional(math::color(1, 1, 0.8f), 10.f));
+            sun.add_component<transform>(position(10, 10, 10), rotation::lookat(math::vec3(1, 1, -1), math::vec3::zero), scale());
+        }
+
+        createProcess<&PhysicsTestSystem::colliderDraw>("Physics",0.02f);
+
+        //quickhullTestScene();
+        //BoxStackScene();
```
LeonBrands: remove commented out code

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+        stabilityComparisonScene();
+    }
+
+    void PhysicsTestSystem::colliderDraw(time::span dt)
+    {
+        //drawPhysicsColliders();
```
LeonBrands: remove commented out code

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+            auto populateMesh = [&calculateUV,&vertices,&normals,&uvs](HalfEdgeEdge* edge)
+            {
+                math::vec3 normal = edge->face->normal;
+               
+                vertices.push_back(edge->edgePosition);
+                uvs.push_back(calculateUV(edge->edgePosition));
+                normals.push_back(normal);
+
+                //log::debug("")
```
LeonBrands: remove commented out code

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+        (math::vec3(10.0f, 5.0f, -0.8f), hammerH, wireFrameH);
+
+        ////suzanne
+        createQuickhullTestObject
+        (math::vec3(15.0f, 5.0f, -0.8f), suzzaneH, wireFrameH);
+
+        ////ohio teapot
+        createQuickhullTestObject
+        (math::vec3(20.0f, 5.0f, -0.5f), teapotH, wireFrameH,elongatedBlockInertia);
+
```
LeonBrands, line 340: fix double newline

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+        litCount++;
+    }
+
+    void PhysicsTestSystem::quickhullTestScene()
+    {
+        math::mat3 elongatedBlockInertia = math::mat3(math::vec3(6.0f, 0, 0), math::vec3(0.0f, 3.0f, 0), math::vec3(0, 0, 6.0f));
+
+        //cube
+        createQuickhullTestObject
```
LeonBrands: formatting for these function calls seem weird

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+        (math::vec3(0,5.0f, -0.8f),cubeH, wireFrameH);
+
+        //cup
```
LeonBrands: the comments here are a bit excessive, the createQuickhullTestObject function calls are one-liners and the variable name (cubeH, hammerH, etc) is sufficiently descriptive

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+                        debug::drawLine(worldStart + shift, worldEnd + shift, usedColor, 2.0f, 0.0f, useDepth);
+
+                    } while (initialEdge != currentEdge && currentEdge != nullptr);
+                }
+            }
+        }
+    }
+
+    int step = 0;
```
LeonBrands: move to the PhysicsTestSystem class as private variables

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,752 @@
...
+                auto meshFilter = ent.get_component<mesh_filter>();
+
+                //[1] clear colliders list
+                auto& physicsComponent = ent.get_component<physics::physicsComponent>().get();
+
+                physicsComponent.colliders.clear();
+                physicsComponent.constructConvexHullFromVertices(meshFilter.get().shared_mesh->vertices);
+
+                //[4] use collider to generate follower objects
+                //PopulateFollowerList(ent,i);
```
LeonBrands: remove commented out code

applications/sandbox/systems/physics_test_system.hpp
```diff
@@ -0,0 +1,110 @@
+#pragma once
+#include <core/core.hpp>
+#include <rendering/components/renderable.hpp>
+#include <physics/cube_collider_params.hpp>
+using namespace legion;
```
LeonBrands: never use `using namespace` in a header

legion/engine/physics/halfedgeface.cpp
```diff
@@ -0,0 +1,178 @@
...
+        {
+            edges.push_back(edge);
+        };
+
+        forEachEdgeReverse(collectEdges);
+
+        for (size_t i = 0; i < edges.size(); i++)
+        {
+            int nextIndex =( i + 1) % edges.size();
+            int prevIndex = (i - 1) == -1 ? edges.size()-1 : (i - 1);
```
LeonBrands: You're comparing integers of different signedness here, keep in mind that i is of type size_t and thus (i -1) will result in UINT_MAX when it overflows. This particular comparison might work out because the -1 will be implicitly converted to uint and overflow as well, but this assumption that i is capable of having negative numbers is error-prone. Either convert i to int or use -1u for the comparison so that the overflow is intentional

legion/engine/physics/halfedgeface.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+            edges.at(i)->setNextAndPrevEdge(newNext, newPrev);
+        }
+
+        startEdge = edges.at(0);
+    }
+
+    void HalfEdgeFace::DEBUG_DrawFace(const math::mat4& transform,const math::color& debugColor,float time)
+    {
+
```
LeonBrands: remove excessive newlines

legion/engine/physics/halfedgeface.cpp
```diff
@@ -0,0 +1,178 @@
...
+        {
+            if (edge->pairingEdge && edge->pairingEdge && edge->pairingEdge->pairingEdge == edge)
+            {
+                edge->pairingEdge->pairingEdge = nullptr;
+            }
+            delete edge;
+        };
+
+        forEachEdge(deleteFunc);
+
```
LeonBrands: remove excessive newline

legion/engine/physics/physics_statics.hpp
```diff
@@ -20,12 +21,12 @@ namespace legion::physics
 
         //---------------------------------------------------------------- Collision Detection ----------------------------------------------------------------------------//
 
-        /** @brief Given 2 ConvexCollider and their respective transforms, checks if 
-        * the colliders are colliding. The result is recorded in the physics_manifold
-        */
-        static void DetectConvexConvexCollision(ConvexCollider* convexA, ConvexCollider* convexB
-            , const math::mat4& transformA, const math::mat4& transformB,
-            ConvexConvexCollisionInfo& outCollisionInfo,  physics_manifold& manifold);
+        ///** @brief Given 2 ConvexCollider and their respective transforms, checks if
```
LeonBrands: remove commented out code

LeonBrands (changes requested):

applications/sandbox/systems/physics_test_system.hpp
GlynLeine: i'd suggest using the `lgn` namespace alias for everything that would require the namespace without this line

legion/engine/physics/colliders/physicscollider.hpp
```diff
@@ -16,27 +16,27 @@ namespace legion::physics
...
+            for (auto&& convergenceId : convergenceIdentifiers)
             {
-                if (converganceId->refColliderID == contact.refCollider->GetColliderID())
+                if (convergenceId->refColliderID == contact.refCollider->GetColliderID())
                 {
-                    if (converganceId->IsEqual(contact))
+                    if (convergenceId->IsEqual(contact))
                     {
-                        converganceId->CopyLambdasToContact(contact);
+                        convergenceId->CopyLambdasToContact(contact);
```
GlynLeine: are you trying to move the unique pointers out? i don't think this is legal code?

legion/engine/physics/data/collider_face_to_vert.cpp
```diff
@@ -0,0 +1,34 @@
...
+            if (distance > largestDist)
+            {
+                largestDist = distance;
+                mostDistantVert = &vert;
+            }
+        }
+
+        return std::make_pair(*mostDistantVert,largestDist);
+    }
+    void ColliderFaceToVert::populateVectorWithVerts(std::vector<math::vec3>& vertVector)
```
GlynLeine: nitpick, but i generally like keeping a line of whitespace between fucntions

legion/engine/physics/data/collider_face_to_vert.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace legion::physics
+{
+    class ColliderFaceToVert
+    {
+    public:
+
+        std::vector<math::vec3> outsideVerts;
+        HalfEdgeFace* face = nullptr;
+
+        ColliderFaceToVert(HalfEdgeFace* pFace) : face(pFace)
```
GlynLeine: technically if this is a non owning pointer you could use the `core::pointer<T>` wrapper, it doesn't allow you to call new or delete on it. but that's also personal preference. idm keeping it like this either

legion/engine/physics/data/convexconvexpenetrationquery.cpp
```diff
@@ -25,15 +24,14 @@ namespace legion::physics
...
-
             float currentDotResult = math::dot(-normal, worldFaceNormal);
             if (currentDotResult > largestDotResult)
             {
                 largestDotResult = currentDotResult;
                 incFace = face;
             }
         }
-
+
```
GlynLeine: have you checked if this line ending change means it's not `LF` instead of `CRLF`?

legion/engine/physics/data/physics_manifold.hpp
```diff
@@ -21,17 +21,17 @@ namespace legion::physics
...
-        physics::physicsComponent* physicsCompB;
+        physics::physicsComponent* physicsCompA = nullptr;
+        physics::physicsComponent* physicsCompB = nullptr;
 
-        physics::rigidbody* rigidbodyA;
-        physics::rigidbody* rigidbodyB;
+        physics::rigidbody* rigidbodyA = nullptr;
+        physics::rigidbody* rigidbodyB = nullptr;
 
         std::unique_ptr<PenetrationQuery> penetrationInformation;
```
GlynLeine, line 38: commented out code under this line

legion/engine/physics/data/physics_manifold.hpp
```diff
@@ -76,20 +76,20 @@ namespace legion::physics
...
             if (idHA)
             {
-                str1 = idHA.read().id;
+                str1 = idHA.get().id;
             }
 
             if (idHB)
             {
-                str2 = idHB.read().id;
+                str2 = idHB.get().id;
```
GlynLeine: psst, you can use the `->` operator on the new component handles

legion/engine/physics/data/physics_manifold_precursor.hpp
```diff
@@ -10,13 +10,13 @@ namespace legion::physics
...
-        physicsComponent* physicsComp;
+        physicsComponent* physicsComp = nullptr;
         id_type id;
-        ecs::entity_handle entity;
+        ecs::entity entity;
 
         physics_manifold_precursor() = default;
 
-        physics_manifold_precursor(math::mat4 pWorldTransform, physicsComponent* pPhysicsComp, id_type precursorID, ecs::entity_handle entity) :
+        physics_manifold_precursor(math::mat4 pWorldTransform, physicsComponent* pPhysicsComp, id_type precursorID, ecs::entity entity) :
```
GlynLeine, line 20: it's generally not recommended to store component pointers, they are not guaranteed to stay valid between frames. but this is fine if this never persists between frames. but in that case maybe add a comment?

legion/engine/physics/data/physics_manifold.hpp
```diff
@@ -21,17 +21,17 @@ namespace legion::physics
...
 
-        physics::physicsComponent* physicsCompA;
-        physics::physicsComponent* physicsCompB;
+        physics::physicsComponent* physicsCompA = nullptr;
+        physics::physicsComponent* physicsCompB = nullptr;
 
-        physics::rigidbody* rigidbodyA;
-        physics::rigidbody* rigidbodyB;
+        physics::rigidbody* rigidbodyA = nullptr;
+        physics::rigidbody* rigidbodyB = nullptr;
```
GlynLeine, line 35: it's generally not recommended to store component pointers, they are not guaranteed to stay valid between frames. but this is fine if this never persists between frames. but in that case maybe add a comment?

legion/engine/physics/halfedgeedge.cpp
```diff
@@ -0,0 +1,236 @@
...
+    {
+        return isVertexVisible(vert,epsilon) && !pairingEdge->isVertexVisible(vert,epsilon);
+    }
+
+    void HalfEdgeEdge::DEBUG_drawEdge(const math::mat4& transform, const math::color& debugColor, float time, float width)
+    {
+        math::vec3 worldStart = transform * math::vec4(edgePosition, 1);
+        math::vec3 worldEnd = transform * math::vec4(nextEdge->edgePosition, 1);
+
+        debug::user_projectDrawLine(worldStart, worldEnd, debugColor, width, time, true);
```
GlynLeine: `debug::user_projectDrawLine` is not guaranteed to exist, use `debug::drawLine()`. `debug::user_projectDrawLine` won't exist without sandbox. also in the other instances of the function call in this file

legion/engine/physics/halfedgeface.cpp
GlynLeine: I'd even just say change it to: `((i  == 0ull ? (edges.size() - 1ull) : (i - 1ull));`

legion/engine/physics/physics_statics.cpp
```diff
@@ -60,18 +62,40 @@ namespace legion::physics
...
+    {
+        float largestDistanceInDirection = std::numeric_limits<float>::lowest();
+        planePosition = math::inverse(colliderTransform) * math::vec4(planePosition, 1);
+        direction = math::inverse(colliderTransform) * math::vec4(direction, 0);
+
+        const auto& vertices = collider->GetVertices();
+
+        for (int vertexIndex = 0 ; vertexIndex < vertices.size(); ++vertexIndex)
+        {
+            math::vec3 transformedVert = math::vec4(vertices.at(vertexIndex), 1);
```
GlynLeine: you're doing a signed unsigned comparison, and `vertexIndex` gets implictly casted to a `size_type` every iteration. this is not costless.

legion/engine/physics/physics_statics.cpp
```diff
@@ -105,22 +128,21 @@ namespace legion::physics
...
+                            math::vec3 edgeBDirection = transformB * math::vec4(edgeB->getRobustEdgeDirection(), 0);
 
-                            edgeADirection = math::normalize(edgeADirection);
-                            edgeBDirection = math::normalize(edgeBDirection);
+                            edgeADirection = math::normalize( edgeADirection );
+                            edgeBDirection = math::normalize( edgeBDirection );
 
-                            //get the seperating axis
-                            math::vec3 seperatingAxis = math::cross(edgeADirection, edgeBDirection);
+                            static float angleThres = math::cos(math::deg2rad(3.0f));
```
GlynLeine: I have a feeling this probably should be in a more public place than locally inside this function? also local static variables aren't 0 cost. they have a thread-safe initialization check around them that will be called every time this function is called.

legion/engine/physics/physics_statics.cpp
```diff
@@ -341,5 +287,728 @@ namespace legion::physics
...
+        math::vec3& secondDistant = supportVertices.at(1);
+
+        //Iterate through support vertices to find the combination of vertices that represent the 2 most distant points
+        for (int i = 0; i < supportVertices.size(); i++)
+        {
+            math::vec3& first = supportVertices.at(i);
+
+            for (int j = i+1; j < supportVertices.size()-1; j++)
+            {
+                math::vec3& second = supportVertices.at(j);
```
GlynLeine: same as before, signed unsigned compare, and `at(size_t)` will cause a cast to `size_type`

legion/engine/physics/physics_statics.cpp
```diff
@@ -341,5 +287,728 @@ namespace legion::physics
...
+
+        for (auto face : faces)
+        {
+            if (initialHorizon) { break; }
+
+            face->forEachEdge(findFirstHorizon);
+            
+        }
+
+        assert(initialHorizon);
```
GlynLeine, line 769: at some point, maybe with the refactor, this should probably get properly handled using `common::result` instead of crashing the entire engine

legion/engine/physics/physics_statics.cpp
```diff
@@ -341,5 +287,728 @@ namespace legion::physics
...
+        HalfEdgeEdge* connectingEdge, float scalingEpsilon,  math::vec3& outNormal,int skipCount)
+    {
+        //each face will at least have 3 vertices
+        std::vector<math::vec3> NewellPolygon;
+        NewellPolygon.reserve(6);
+
+        auto firstOriginal = first->startEdge;
+        auto secondOriginal = second->startEdge;
+
+        assert(first == connectingEdge->face);
```
GlynLeine, line 975: same as mentioned before about `common::result`

legion/engine/physics/systems/physicssystem.cpp
```diff
@@ -3,24 +3,40 @@
...
+        size_type currentInterval = 0;
 
-        m_broadPhase = std::make_unique<BroadphaseUniformGridNoCaching>(math::vec3(2, 2, 2));
+        while (m_accumulator > m_timeStep && currentInterval < m_maxInterval)
+        {
+            m_accumulator -= m_timeStep;
+            fixedUpdate(m_timeStep);
 
+            currentInterval++;
+        }
```
GlynLeine: wait why exactly did we stop using the fixed interval process?

legion/engine/physics/systems/physicssystem.hpp
```diff
@@ -235,11 +183,11 @@ namespace legion::physics
         void integrateRigidbodies(std::vector<byte>& hasRigidBodies, ecs::component_container<rigidbody>& rigidbodies, float deltaTime)
         {
             OPTICK_EVENT();
-            m_scheduler->queueJobs(manifoldPrecursorQuery.size(), [&]() {
+            queueJobs(manifoldPrecursorQuery.size(), [&]() {
                 if (!hasRigidBodies[async::this_job::get_id()])
                     return;
 
-                auto& rb = rigidbodies[async::this_job::get_id()];
+                rigidbody& rb = rigidbodies[async::this_job::get_id()].get();
```
GlynLeine: fyi, these all work:
```cpp
rigidbody& rb = rigidbodies[async::this_job::get_id()];
```
or
```cpp
auto& rb = *rigidbodies[async::this_job::get_id()];
```
or
```cpp
auto& rb = rigidbodies[async::this_job::get_id()].get();
```

legion/engine/physics/colliders/physicscollider.hpp
Developer-The-Great: there is no move being done here, some of the floats of  Physics_Contact are being copied to ConvergenceIdentifier

GlynLeine: this this should still be an `auto&` since i don't recall calling functions or writing data to a rvalue reference being generally considered well defined. since rvalues are meant to be temporary values. (right hand side values)

Developer-The-Great: ah did not notice the '&&' will fix that

legion/engine/physics/data/physics_manifold.hpp
Developer-The-Great: This doesn't persist between frames, will add that comment mentioning that

legion/engine/physics/data/physics_manifold_precursor.hpp
Developer-The-Great: this never persist between frames, will add that comment

legion/engine/physics/halfedgeface.cpp
Developer-The-Great: I'll convert it i to int since it would be much more readable

GlynLeine: actually that was the wrong thing to take away here, since `.size()` will always return a `size_type` anyways. turning i into an int will just cause signed unsigned comparison, and implicit casts from `size_type` to `int`. actually `nextIndex` and `prevIndex` should also be `size_type`.

generally if you're dealing with indices and sizes and amounts, you want to be using `size_type`

legion/engine/physics/physics_statics.cpp
Developer-The-Great: i do recall that I wanted this to be a constexpr value in physics::constants, but this wasn't possible because there is no constexpr version of math::cos. I could just declare the literal value of math::cos(math::deg2rad(3.0f)) in physics::constants, is that sufficient?

GlynLeine: yeah if you can't make it `constexpr`, it's fine to make it `static const`. it won't be generated on compile time, but it will be generated at startup.

legion/engine/physics/data/collider_face_to_vert.hpp
Developer-The-Great: image

core::pointer currently doesn't have a constructor that takes in the T* in question. Which means I cant value initialize my pointers to nullptr. Will keep using raw pointers in this specific situation for now.

legion/engine/physics/physics_statics.cpp
Developer-The-Great: would you prefer me to handle this in this pr or in another separate pr?

legion/engine/physics/systems/physicssystem.cpp
Developer-The-Great: Initially, this was because there seemed to be something wrong with createProcess causing it to call fixedUpdate in the wrong intervals. I looked at this section of the code again and did some tests, apparently, it was something entirely unrelated (my bad!). But yeah good catch, will revert this.

legion/engine/physics/systems/physicssystem.hpp
Developer-The-Great: good to know!

legion/engine/physics/data/collider_face_to_vert.hpp
GlynLeine: that's becasue it's a trivial type:
```cpp
core::pointer<HalfEdgeFace> face2 = { nullptr };
```

legion/engine/physics/physics_statics.cpp
GlynLeine: this will probably require a bigger architectural change, just like the use of raw memory management throughout physics. I think this should come in with whatever PR will come with that one.

legion/engine/physics/systems/physicssystem.cpp
GlynLeine: If you would like to change the behaviour of fixed interval process, please let me know. physics was the main motivation behind it's existence. (and also the only that uses it rn) So it's fine to say add the maximum updates per frame that you added here.

GlynLeine: or alternatively you could ofc add it yourself

Developer-The-Great: I might change my mind about this later on, but I think I prefer using createProcess since this essentially achieves what we want to do (update physics in a fixed interval) with comparitively less code.

LeonBrands (approved):
changes lgtm. I'd wait for a review from Glyn before merging though

legion/engine/physics/systems/physicssystem.cpp
GlynLeine: no what i meant is: you're allowed to go into system and scheduling to change how fixed interval process works. admittedly i didn't spend all too much research into how it should or shouldn't behave.

Developer-The-Great: alright, but this is not something I need to do right now since it works as intended for now

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,729 @@
...
+        for (auto ent : objectToFollowQuery)
+        {
+            auto& objToFollow = ent.get_component<ObjectToFollow>().get();
+
+            auto [posH,rotH,scaleH] = ent.get_component<transform>();
+            auto [posH2, rotH2, scaleH2] = objToFollow.ent.get_component<transform>();
+
+            posH = posH2.get();
+            rotH = rotH2.get();
+            scaleH = scaleH2.get();
```
GlynLeine, line 111: I get that this is just a test system, so i don't rly care tbh. but keep in mind that any of these handles might be invalid. and that will cause an exception.

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,729 @@
...
+        //using vertices of convex hull, create a rendering mesh out of it
+        auto convexCollider = std::dynamic_pointer_cast<ConvexCollider>(newPhysicsComponent.colliders.at(0));
+
+        mesh newMesh;
+
+        std::vector<uint>& indices = newMesh.indices;
+        std::vector<math::vec3>& vertices = newMesh.vertices;
+        std::vector<math::vec2>& uvs = newMesh.uvs;
+        std::vector<math::vec3>& normals = newMesh.normals;
+        std::vector<math::color>& colors = newMesh.colors;
```
GlynLeine, line 205: tbh i would use auto& here, but up to you XD

applications/sandbox/systems/physics_test_system.cpp
```diff
@@ -0,0 +1,729 @@
...
+        newSubMesh.materialIndex = 0;
+
+        newMesh.submeshes.push_back(newSubMesh);
+
+        static int procCount = 0;
+        std::string name = std::string("QuickhullMesh") + std::to_string(procCount);
+
+        procCount++;
+
+        core::assets::asset<mesh> meshAsset = core::assets::AssetCache<mesh>::create(name,newMesh);
```
GlynLeine, line 291: yeah i'd recommend getting used to using lot's of auto, but ig this is prolly pretty old code since i do see you using auto everywhere else XD

GlynLeine (approved):
lgtm!

PR #399, opened 2022-02-16
Subfeature/add physx include and libs

Developer-The-Great:
- added physx includes in the deps folder
- add physx lib and dll files
- add a physx system that initializes physx globals and creates an empty PxScene


legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -0,0 +1,60 @@
...
+    PxPvd* gPvd = NULL;
+
+    PxPhysics* gPhysics = NULL;
+
+    PxDefaultCpuDispatcher* gDispatcher = NULL;
+
+    //PhysX PVD Debugger Related
+    constexpr char* pvdHost = "127.0.0.1";
+    constexpr size_type defaultPVDListeningPort = 5425;
+    constexpr size_type defaultPVDHostTimeout = 10;
```
GlynLeine: I'd personally rather not have these as floating globals, I'd be fine having them as part of some static struct, but rather not as floating globals. Also for their initialization, it might be possible for multiple systems or engine instances to exist, and these probably should be shut down gracefully with the system when the system is shut down or restarted

legion/engine/physics/physx/systems/physx_physics_system.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+        void fixedUpdate(time::time_span<fast_time> deltaTime);
+
+    private:
+
+        void setupPhysXGlobalVariables();
+
+        void setupDefaultScene();
+
+        float m_timeStep = 0.02f;
```
GlynLeine: i'd say this could be a static constexpr value

legion/engine/physics/physx/systems/physx_physics_system.cpp
Developer-The-Great: yeah most of these don't need to be globals. With the exception of PxPhysics as stated in the docs here.
image
Will change all the other globals into statics tho.

LeonBrands: Are you sure you're reading the documentation right here? I believe name mangling is only really relevant for functions so to me this seems like they're referring to the function PxCreatePhysics() and not the resulting value

LeonBrands:
What's the license situation like for PhysX?

legion/engine/physics/physx/systems/physx_physics_system.cpp
Developer-The-Great: I don't really know much about name mangling but this mentions that it can happen to variables. https://www.ibm.com/docs/en/i/7.2?topic=linkage-name-mangling-c-only
I guess I could just make it static and simply revert it back to a global when it causes problems later on

Developer-The-Great:
> What's the license situation like for PhysX?

From what I understand, I think as long as I keep the copyright notice in the includes we shouldn't have any issues? https://github.com/NVIDIAGameWorks/PhysX/blob/4.1/LICENSE.md

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,56 +5,96 @@ namespace legion::physics
...
-        setupPhysXGlobalVariables();
+        lazyInitPhysXVariables();
         setupDefaultScene();
-        
+
         createProcess<&PhysXPhysicsSystem::fixedUpdate>("Physics", m_timeStep);
     }
 
-    void PhysXPhysicsSystem::fixedUpdate(time::time_span<fast_time> deltaTime)
+    PhysXPhysicsSystem::~PhysXPhysicsSystem()
```
GlynLeine: this should prolly be the `shutdown` function instead, since systems might be restarted without being deconstructed

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,56 +5,96 @@ namespace legion::physics
...
+    void PhysXPhysicsSystem::releasePhysXVariables()
+    {
+        sPhysicsStatics.dispatcher->release();
+        gPhysics->release();
+
+        if (sPhysicsStatics.pvd)
+        {
+            PxPvdTransport* transport = sPhysicsStatics.pvd->getTransport();
+            sPhysicsStatics.pvd->release();
+            sPhysicsStatics.pvd = nullptr;
```
GlynLeine: is there a specific reason for things being called `s...` or `g...` instead of the normal `m_...`?

legion/engine/physics/physx/systems/physx_physics_system.hpp
```diff
@@ -12,18 +17,39 @@ namespace legion::physics
...
 
-        void setupPhysXGlobalVariables();
+        struct physXStatics
+        {
+            bool m_isInit = false;
+
+            physx::PxFoundation* foundation = nullptr;
+            physx::PxPvd* pvd = nullptr;
+            physx::PxDefaultCpuDispatcher* dispatcher = nullptr;
+        };
```
GlynLeine: if you forward declare `physx_statics` here then you can define the definition in the .cpp file. since i don't think anything in this header needs to know about these implementation details.

```
// hpp
class PhysxSystem final : public System<PhysxSystem>
{
    struct physx_statics;
};
```
```
// cpp
struct PhysxSystem::physx_statics
{
    static size_type systemInstances;
    static physx::PxFoundation* foundation = nullptr;
    static physx::PxPvd* pvd = nullptr;
    static physx::PxDefaultCpuDispatcher* dispatcher = nullptr;
}
```

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,56 +5,96 @@ namespace legion::physics
...
+            releasePhysXVariables();
+        }
     }
 
-    void PhysXPhysicsSystem::setupPhysXGlobalVariables()
+    void PhysXPhysicsSystem::lazyInitPhysXVariables()
     {
-        gFoundation = PxCreateFoundation(PX_PHYSICS_VERSION, gAllocator, gErrorCallback);
+        static PxDefaultAllocator sDefaultAllocator;
+        static PxDefaultErrorCallback sDefaultErrorCallback;
```
GlynLeine: why do we have these two if the system itself also already has these?

legion/engine/physics/physx/systems/physx_physics_system.hpp
```diff
@@ -12,18 +17,39 @@ namespace legion::physics
     {
     public:
 
+        PhysXPhysicsSystem();
```
GlynLeine: I can't find the definition of this constructor? also i don't think it should be necessary to have any defined constructors or destructors?

legion/engine/physics/physx/systems/physx_physics_system.cpp
Developer-The-Great: Mostly because I assumed m_ is only for private member variables of classes, I thought we could use something else for statics and globals. I can easily change this of course.

legion/engine/physics/physx/systems/physx_physics_system.hpp
Developer-The-Great: its on the cpp file on line 18, it adds the instance counter. But I guess this can be done on setup instead

legion/engine/physics/physx/systems/physx_physics_system.cpp
Developer-The-Great: mostly because I wanted all physx systems to point to the same instance of these classes, this will be changed by putting it in the new physXStatics struct

legion/engine/physics/physx/systems/physx_physics_system.cpp
Developer-The-Great: Just tested it, no issues with making it static

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,91 +5,98 @@ namespace legion::physics
 {
     using namespace physx;
 
-    PhysXPhysicsSystem::physXStatics PhysXPhysicsSystem::sPhysicsStatics;
-    size_type PhysXPhysicsSystem::sSelfInstanceCounter = 0;
+    struct physXStatics
+    {
+        static size_type sSelfInstanceCounter;
```
GlynLeine: does not adhere to naming conventions

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,91 +5,98 @@ namespace legion::physics
...
+        static PxDefaultErrorCallback defaultErrorCallback;
+    };
 
-    PxPhysics* gPhysics = nullptr;
+    size_type physXStatics::sSelfInstanceCounter = 0;
 
+    PxFoundation* physXStatics::foundation = nullptr;
+    PxPvd* physXStatics::pvd = nullptr;
+    PxDefaultCpuDispatcher* physXStatics::dispatcher = nullptr;
+    PxPhysics* physXStatics::phyxSDK = nullptr;
```
GlynLeine: if the struct only exist in the cpp file, then you might as well do:
```
struct foo
{
    inline static BarType* bar = nullptr;
};
```

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,91 +5,98 @@ namespace legion::physics
...
         m_physxScene->release();
 
-        sSelfInstanceCounter--;
+        physXStatics::sSelfInstanceCounter--;
 
-        if (sSelfInstanceCounter == 0)
+        if (physXStatics::sSelfInstanceCounter == 0)
         {
             releasePhysXVariables();
         }
```
GlynLeine: initialization and uninitialization should probably be protected with a lock

legion/engine/physics/physx/systems/physx_physics_system.hpp
```diff
@@ -31,5 +31,7 @@ namespace legion::physics
         static constexpr float m_timeStep = 0.02f;
 
         physx::PxScene* m_physxScene;
+
+        std::mutex m_setupShutdownMutex;
```
GlynLeine: If this lock is not static between all instances of the system then it doesn't do anything XD

GlynLeine: also using async::spinlock and async::lock_guard will be a little nicer and technically a bit faster, although i doubt that matters much since it's startup and shutdown.
```
{
    async::lock_guard guard(PS::setupShutdownLock);
    // Do stuff
}
```

legion/engine/physics/physx/systems/physx_physics_system.cpp
```diff
@@ -5,38 +5,32 @@ namespace legion::physics
...
 
     void PhysXPhysicsSystem::setup()
     {
-        physXStatics::sSelfInstanceCounter++;
+        const std::lock_guard<std::mutex> setupLock(m_setupShutdownMutex);
 
+        PS::selfInstanceCounter++;
+        
         lazyInitPhysXVariables();
         setupDefaultScene();
```
GlynLeine, line 35: also i don't think createProcess should be called within the critical section of the lock

PR #405, opened 2022-03-03
Task/core build system

LeonBrands:
Sets up the basic build system for rythe engine/editor - The current system detects modules and allows enabling them through CMake flags. If a module is enabled, a matching editor module will be attempted to be enabled as well.

Header files and library files are copied to the appropriate include/ and lib/ folders.

This version contains no applications, tests, or docs of any form yet.


LeonBrands:
note that there are no current protections on develop - this PR is not a git-flow requirement, but exists nonetheless for receiving feedback on the current status of the build system. Do not merge after a single accepted review, I will do so when all (or sufficient) reviews are done

Ragingram2 (approved):

CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+
+set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+
+# Convenience directory variables, usable in subdirs
+set(RYTHE_DIR_ROOT ${CMAKE_CURRENT_SOURCE_DIR})
+set(RYTHE_DIR_RYTHE ${CMAKE_CURRENT_SOURCE_DIR}/rythe)
+set(RYTHE_DIR_ENGINE ${CMAKE_CURRENT_SOURCE_DIR}/rythe/engine)
+set(RYTHE_DIR_ENGINE_MODULES ${CMAKE_CURRENT_SOURCE_DIR}/rythe/engine/modules)
+set(RYTHE_DIR_EDITOR ${CMAKE_CURRENT_SOURCE_DIR}/rythe/engine)
```
GlynLeine: `RYTHE_DIR_EDITOR` is `.../rythe/engine`?

CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+        set(RYTHE_COMPILER_FLAGS /WX ${RYTHE_COMPILER_FLAGS})
+    endif()
+
+	if (${RYTHE_BUILD_OPTION_ASAN}})
+		set(RYTHE_COMPILER_FLAGS /fsanitize=address ${RYTHE_COMPILER_FLAGS})
+	endif()
+endif()
+
+# Handle static/shared libs
+# TODO (Leon): This could be more elegant and we could potentially allow for dynamic libraries.
```
GlynLeine: dynamic libs is something that should prolly be discussed at some point in the future since they might require more work. I have been thinking about maybe keeping core as a static lib and having something like a default runner that loads a game DLL and calls the reportModules function from the DLL.

but dynamic libraries of the other default modules is not something that has been explored much more than the initial decision of going static for various reasons we had back then. (link time optimization and not having to worry about DLL security)

GlynLeine:
> note that there are no current protections on develop - this PR is not a git-flow requirement, but exists nonetheless for receiving feedback on the current status of the build system. Do not merge after a single accepted review, I will do so when all (or sufficient) reviews are done

would you prefer working on this through the default git use and PR workflow? since I removed all protections for any hard changed you had to make or cases where you had to force git to do what you want. but i feel like we've gotten past that by now? I could setup the branch protection back up again.

LeonBrands:
@GlynLeine leave it off for now, I might need to do some more work where gitflow will just get in the way. Am kind of selectively picking things that I want reviewed right now

CMakeLists.txt
LeonBrands: oops!

CMakeLists.txt
LeonBrands: if we have no near-future plans to make stuff dynamic then I can probably just remove the todo for now

GlynLeine: I'd say prolly not near-future there's too many other things that need doing first. and allowing users to choose whether to go dynamic or static will also take a lot of time tbh. because we would need to make sure everything is exported and imported correctly. specifically static member variables might be problematic iirc

GlynLeine (approved):

PR #406, opened 2022-03-05
Task/build applications

LeonBrands:
Automatically adds in applications from modules when the modules are enabled.

This PR also fixes an issue where submodules were added to the linker even if they didn't have c or cpp files (and thus wouldn't generate a static library)

### DIFF
--- a/cmake/copy_module_output.cmake
+++ b/cmake/copy_module_output.cmake
@@ -19,24 +19,26 @@ function(copy_module_output targetName targetDir)
 		make_directory ${RYTHE_DIR_OUTPUT_LIBS}/Release
 	)
 
-	# Copy debug library
-	# Note: || (exit 0) makes sure these commands succeed even if no library file was generated.
-	add_custom_command(TARGET ${targetName} 
-		POST_BUILD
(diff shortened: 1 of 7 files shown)


cmake/copy_module_output.cmake
```diff
@@ -19,24 +19,26 @@ function(copy_module_output targetName targetDir)
...
-	add_custom_command(TARGET ${targetName} 
-		POST_BUILD
-		COMMAND $<$<CONFIG:Debug>:${CMAKE_COMMAND}> -E 
-		copy 
-			$<TARGET_FILE:${targetName}>
-			"${RYTHE_DIR_OUTPUT_LIBS}/Debug/" || (exit 0)
-	)
+	if (${MODULE_INFO_HAS_SOURCES})
+		# Copy debug library
+		# Note: || (exit 0) makes sure these commands succeed even if no library file was generated.
```
GlynLeine, line 24: in which cases could that actually happen?

LeonBrands: Not many - it's possible for it to occur when the source itself has errors, in which case no .lib file gets generated, but the post build command could still try to run, which'll result in a difficult to understand error so I'd rather hide it I think

PR #407, opened 2022-03-10
added in build & analyze actions

LeonBrands:
analyze actions currently fail because there's no code. they were tested on a local branch with submodules added in and they do work fine.

Will be automatically fixed later once the modules are set up properly

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -1,0 +1,68 @@
+# GH Action: Analyze
+# Runs various code analysis tools on the codebase
+name: analyze
+on: [push, pull_request]
+  
+jobs:
+  clang-format:
+    if: false # Disabled until a .clang-format style guide is established
+    name: clang-format
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Download submodules
+      run: git submodule update --init --recursive
+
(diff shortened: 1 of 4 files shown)


.github/workflows/analyze.yml
```diff
@@ -0,0 +1,68 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Download submodules
+      run: git submodule update --init --recursive
+
+    - name: Get cppcheck
+      run: DEBIAN_FRONTEND=noninteractive sudo apt-get update -yqq --allow-releaseinfo-change -o::Dpkg::Use-Pty=0 && DEBIAN_FRONTEND=noninteractive sudo apt-get install -yqq -o::Dpkg::Use-Pty=0 cppcheck
+
+    - name: cppcheck
+      run: cppcheck rythe/ --inline-suppr --suppress=syntaxError --xml --error-exitcode=1 2>&1 | python3 tools/cppcheck2github.py
```
GlynLeine, line 38: why suppress syntax errors?

LeonBrands: I don't actually know, these were the parameters set in the analyze action previously! :p

GlynLeine: welp ig we should ask @Algo-ryth-mix then XD

GlynLeine (approved):

PR #409, opened 2022-12-27
Preview only PR

GlynLeine:
PR only for @LeonBrands to see my changes on the CMake


CMakeLists.txt
```diff
@@ -116,14 +116,68 @@ endif()
 
 # Add Rythe
 set(RYTHE_ENABLED_MODULE_NAMES "")
+set(RYTHE_ENABLED_APPLICATION_NAMES "")
 set(RYTHE_ENABLED_MODULE_PATHS "")
 add_subdirectory(${RYTHE_DIR_RYTHE})
 
 # The engine is ready - add applications
-if (${RYTHE_BUILD_APPLICATIONS})
+if ((${RYTHE_BUILD_APPLICATIONS}))
```
LeonBrands: AFAIK this is unnecessary?

CMakeLists.txt
```diff
@@ -116,14 +116,68 @@ endif()
...
+
+# Add config project
+set(configure_command_args "")
+list(APPEND configure_command_args "\"${RYTHE_DIR_ROOT}\"")
+list(APPEND configure_command_args "-G \"${CMAKE_GENERATOR}\"")
+list(APPEND configure_command_args "-S \"${RYTHE_DIR_ROOT}\"")
+list(APPEND configure_command_args "-B \"${CMAKE_BINARY_DIR}\"")
+list(APPEND configure_command_args -T ClangCL)
+list(APPEND configure_command_args -D RYTHE_FETCH_SUBMODULES=OFF)
+
```
LeonBrands: Can you not just do `list(APPEND configure_command_args -D RYTHE_BUILD_APPLICATIONS=${RYTHE_BUILD_APPLICATIONS})` ?

(and the same for ASAN, WERROR, ENABLE_ALL_MODULES) ?

CMakeLists.txt
```diff
@@ -116,14 +116,68 @@ endif()
...
+foreach(application_name ${RYTHE_ENABLED_APPLICATION_NAMES})
+	add_dependencies(${application_name} configure)
+endforeach()
+
+# FOLDER THINGS
+file(GLOB files ${RYTHE_DIR_RYTHE}/${RYTHE_FILE_TYPES})
+foreach(file ${files})
+	set_target_properties(${file} PROPERTIES FOLDER ${file})
+endforeach()
```
LeonBrands: Missing newline at EOF

CMakeLists.txt
```diff
@@ -116,14 +116,68 @@ endif()
...
+
+foreach(module_name ${RYTHE_ENABLED_MODULE_NAMES})
+	add_dependencies(${module_name} configure)
+endforeach()
+
+foreach(application_name ${RYTHE_ENABLED_APPLICATION_NAMES})
+	add_dependencies(${application_name} configure)
+endforeach()
+
+# FOLDER THINGS
```
LeonBrands: Change to a more descriptive comment like "Set the folder names. Use to organize targets in IDEs."

VisualStudio22-All.bat
```diff
@@ -0,0 +1,2 @@
+cmake -E make_directory build
+cmake . -G "Visual Studio 17 2022" -S . -B ./build/ -T ClangCL -D RYTHE_BUILD_APPLICATIONS=ON -D RYTHE_BUILD_OPTION_ASAN=ON -D RYTHE_FORCE_ENABLE_ALL_MODULES=ON
```
LeonBrands: You changed RYTHE_BUILD_OPTION_ASAN to be enabled by default, excessive to enable it again in this command

applications/sandbox/CMakeLists.txt
```diff
@@ -1,11 +1,19 @@
 # Sandbox depends on applications and graphics in this case
 
-if (${RYTHE_MODULE_APPLICATION} AND ${RYTHE_MODULE_GRAPHICS})
+if (${RYTHE_BUILD_APPLICATIONS} AND ${RYTHE_MODULE_GRAPHICS})
```
LeonBrands: This seems incorrect, I think you confused what this if statement was for - When this was written, sandbox depended on the Application module, RYTHE_BUILD_APPLICATIONS isn't that. If sandbox still depends on the application module, we should be checking if it's enabled.

You shouldn't have to check for RYTHE_BUILD_APPLICATIONS here, applications aren't enabled at all if RYTHE_BUILD_APPLICATIONS is disabled, meaning this code wouldn't run in the first place

LeonBrands: It could be interesting to explore a more explicit & generalized dependency system for applications, maybe they could report their dependencies and then the applications cmake could check if they can be run

cmake/add_third_party.cmake
```diff
@@ -35,7 +35,14 @@ macro(rythe_add_third_party)
 
 			foreach(target ${RYTHE_ADD_THIRD_PARTY_FOLDER_TARGETS})
 				if (TARGET ${target})
-					set_target_properties(${target} PROPERTIES FOLDER ${RYTHE_ADD_THIRD_PARTY_FOLDER})
+					string(REGEX REPLACE "^third_party" "4 - third party" filter ${RYTHE_ADD_THIRD_PARTY_FOLDER})
```
LeonBrands: Is this a quick fix/hack or the only way of doing this?

LeonBrands (commented):
Couple comments, I didn't review any of the C++ code, only the cmake work

CMakeLists.txt
LeonBrands: Oops, that references the wrong bit of code. This comment is intended  for the lines 141-161, regarding statements like:

```
if ((${RYTHE_BUILD_APPLICATIONS}))
	list(APPEND configure_command_args -D RYTHE_BUILD_APPLICATIONS=ON)
else()
	list(APPEND configure_command_args -D RYTHE_BUILD_APPLICATIONS=OFF)
endif()
```

applications/sandbox/CMakeLists.txt
GlynLeine: No the reason this happened was because i was testing rendering without the application module. it's temporary and it should be:
`if (${RYTHE_BUILD_APPLICATIONS} AND ${RYTHE_MODULE_APPLICATION} AND ${RYTHE_MODULE_GRAPHICS})`

cmake/add_third_party.cmake
GlynLeine: only way i could find..

